PR #9, opened 2023-10-17
Testnet T2 (WIP)

mialbu:
This PR is still work in progress, I'll request reviews once the contract tests are implemented.


contracts/BridgeContract.sol
```diff
@@ -0,0 +1,283 @@
...
+    }
+
+    function verify(
+        MerkleProof calldata _proof
+    )
+        pure
+        private
+        returns (bool)
+    {
+        bytes32 right = sha256(abi.encodePacked(_proof.nonce, _proof.recipient, _proof.amount));
```
cloud8little: why is always sha256 with the leaf node as right?

mialbu:
Replaced by #40

PR #21, opened 2023-11-03
Introduce path in proof verification

### LINKED ISSUE #17: [Bridge Bane] Use path for proof verification
Currently the merkle tree proof is always done from the right to the left. In order to prepare for implementing a claim functionality which requires any leaf to be verified to the root, the contract should use a path in the verification.

- [x] Validators should only sign the concatenation of the top root and the last nonce and not the concatenation of the roots of all deposits.
- [x] Use a bit map to provide a path for the proof verification.

### LINKED ISSUE #18: [Bridge Bane] Implement a claim functionality
The claim functionality should be provided for when the `to` address is a contract, or when the `transfer` returns `false`.

Depends on #17

mialbu:


contracts/BridgeContract.sol
```diff
@@ -114,49 +137,60 @@ contract Bridge {
...
-                    } else {
-                        // Todo: What happens if the transfer fails?
-                        // emit FailedDeposit(proof.nonce, to, proof.amount);
-                    }
-                }
-                // In case the recipient is a contract, no funds are sent. However, the Merkle Tree computation must withstand.
-                // Todo: Discuss if we want to provide refund support in case the recipient is a contract.
-                // emit FailedDeposit(proof.nonce, to, proof.amount);
+            MerkleProof calldata p = _proofs[i];
+            if (isContract(p.to)) {
```
cloud8little: so when the to address is contract, we do not check the proof?

mialbu: Good question! Now, thinking a bit better, we should still verify the proof, just in order that the relayer cannot just provide an address for a nonce that is a contract and not execute it (i.e., not pay for the verification and transfer, and instead move it to the claim mapping) deliberately.

mialbu: See c3c172a

contracts/BridgeContract.sol
```diff
@@ -114,49 +140,106 @@ contract Bridge {
...
-                uint256 transferAmount = toEthDecimals(proof.amount);
-                if (!isContract(to)) {
-                    if (to.send(transferAmount)) {
-                        emit Deposit(proof.nonce, to, proof.amount);
+            MerkleProof calldata p = _proofs[i];
+            if (verify(p)) {
+                if (isContract(p.to)) {
+                    // If the recipient is a contract, the deposit is not transferred to the recipient.
+                    // Instead, the deposit is stored in a mapping and can be claimed individually.
+                    claimableDeposits[p.nonce] = true;
```
cloud8little, line 148: why set the claim table when the recipient address is a contract?

mialbu: The relayer will not pay for unknown code to be executed in the contract's receive or fallback method. Hence, we'll only support EOAs as recipients.

If in this step, the code just ignores the deposit and continues in the loop, we'd be kind of `burning` gas for the depositor. Making it claimable allows the depositor to execute it and pay the execution cost himself.

Do you think we should handle this case differently? I'm open for suggestions.

cloud8little: I am not sure what would be the case when the recipient is a contract, if we add it to the claim table, and the contract actually can not call the claim function, there is some confusion here.

contracts/BridgeContract.sol
```diff
@@ -114,49 +140,106 @@ contract Bridge {
...
         }
     }
 
-    function verify(MerkleProof calldata _proof) private pure returns (bool) {
-        bytes32 right = sha256(
-            abi.encodePacked(_proof.nonce, _proof.to, _proof.amount)
+    // Todo: Add functionality to move deposit to withdrawal without claiming.
+
+    function claim(MerkleProof calldata _proof) external {
+        require(isContract(_proof.to), "Recipient must be an EOA.");
```
cloud8little: guess this would be require _proof.to to be not a contract.

mialbu: > guess this would be require _proof.to to be not a contract.

Yes, just noticed this independently as well.

contracts/BridgeContract.sol
mialbu: > I am not sure what would be the case when the recipient is a contract, if we add it to the claim table, and the contract actually can not call the claim function, there is some confusion here.

In the current design, it doesn't matter who the `msg.sender` is or if the invocation comes from a contract. Anyone could execute a `claim`. The funds in the claim will be sent to the recipient of the deposit that is required for a successful proof.

contracts/BridgeContract.sol
mialbu: Fixed in 1a5e10f

mialbu:
We've continued development in another branch and use a path with a proof there.

PR #24, opened 2023-11-13
add hash tree approach contract tests

cloud8little:


cloud8little:
Done, ready for review @mialbu

test/HashTreeBridgeTest.ts
```diff
@@ -0,0 +1,334 @@
...
+            const new_root = await computeRoot(root1, hashDepositData2);
+            const new_encodeRoot = ethers.solidityPackedKeccak256(["bytes32"], [new_root]);
+            const signatures = await getValidatorSignatures(ethers.getBytes(new_encodeRoot), [1, 2, 3, 4, 5]);
+
+            const tx = await hashTreebridgeContract.connect(relayer).deposit(new_root, signatures, [Depositdata1, Depositdata2]);
+            // check the balance is not change for the recipient address; but the root and depositNonce are both updated
+            await expect(tx).to.changeEtherBalances([hashTreebridgeContract, Depositdata1.to, Depositdata2.to], [0, 0, 0]);
+            expect(await hashTreebridgeContract.depositNonce()).to.equal(Depositdata2.nonce);
+            expect(await hashTreebridgeContract.depositRoot()).to.equal(new_root);
+
```
mialbu, line 145: Can you add an expect statement here, that checks that the two deposits have been added to the claimable mapping?

test/HashTreeBridgeTest.ts
```diff
@@ -0,0 +1,334 @@
...
+        it("Should revert with empty proofs", async function () {
+            const { hashTreebridgeContract, relayer } = await loadFixture(deployBridgeFixture);
+            await expect(hashTreebridgeContract.connect(relayer).deposit(ethers.ZeroHash, [], [])).to.be.revertedWith("At least 1 deposit is required.");
+        });
+
+        // it("Should revert with proofs length greater than 10", async function () {
+        //     const { bridgeContract, relayer } = await loadFixture(deployBridgeFixture);
+        //     const proofs = [proof1, proof2, proof3, proof4, proof5, proof6, proof7, proof8, proof9, proof10, proof11];
+        //     await expect(bridgeContract.connect(relayer).deposit(proofs, [])).to.be.revertedWith("At most 10 proofs are allowed.");
+        // });
```
mialbu: You can uncomment this one, it should pass now.

mialbu: I've set the message to `Too many deposits provided.` because I use a public variable. If we use a const inlined, we should use the number in the message.

test/HashTreeBridgeTest.ts
```diff
@@ -0,0 +1,334 @@
...
+        //     const { hashTreebridgeContract, relayer } = await loadFixture(deployBridgeFixture);
+        //     await fundContract(hashTreebridgeContract, relayer);
+
+        //     const hashDepositData1 = await hashDepositOrWithdrawal(Depositdata1.nonce, Depositdata1.amount, Depositdata1.to);
+        //     const root1 = await computeRoot(ethers.ZeroHash, hashDepositData1);
+        //     const encodeRoot1 = ethers.solidityPackedKeccak256(["bytes32"], [root1]);
+        //     const signatures = await getValidatorSignatures(ethers.getBytes(encodeRoot1), [0, 2, 3, 4]);
+
+        //     await expect(hashTreebridgeContract.connect(relayer).deposit(root1, signatures, [Depositdata1])).to.be.revertedWith("Invalid or insufficient validator signatures.");
+        // });
```
mialbu: Same reasoning here with the message I included: `Invalid number of signatures.`

test/HashTreeBridgeTest.ts
```diff
@@ -0,0 +1,334 @@
...
+
+        it("Should revert when signature verify failed", async function () {
+            const { hashTreebridgeContract, relayer } = await loadFixture(deployBridgeFixture);
+            await fundContract(hashTreebridgeContract, relayer);
+
+            const hashDepositData1 = await hashDepositOrWithdrawal(Depositdata1.nonce, Depositdata1.amount, Depositdata1.to);
+            const root1 = await computeRoot(ethers.ZeroHash, hashDepositData1);
+            const encodeRoot1 = ethers.solidityPackedKeccak256(["bytes32"], [root1]);
+            const signatures = await getValidatorSignatures(ethers.getBytes(encodeRoot1), [0, 2, 3, 4, 5]);
```
mialbu, line 265: Can you please add a comment above this line, mentioning that the provided index `0` refers to the relayer, and thus the signatures will not pass the validator signature verification in the contract.

test/HashTreeBridgeTest.ts
```diff
@@ -0,0 +1,334 @@
...
+            expect(await hashTreebridgeContract.withdrawalRoot()).to.be.equal(hash12);
+            await expect(tx2).to.emit(hashTreebridgeContract, "Withdrawal").withArgs(2, toNeoDecimals(withdrawData2.amount), validator1.address, relayer.address);
+
+        });
+
+        it("withdraw with the wrong amount", async function () {
+            const { hashTreebridgeContract, relayer } = await loadFixture(deployBridgeFixture);
+
+            await expect(hashTreebridgeContract.connect(relayer).withdraw(relayer, { value: ethers.parseEther("0.000000001") })).to.be.revertedWith("Only amounts with 8 non-zero decimals allowed for withdrawal");
+        });
```
mialbu, line 339: Can you add an edge case, where the amount is greater than the minimum deposit, but it has more than 8 decimal places, e.g., `1.000000001` and then another one that should pass with the exact min deposit amount (`1.00000001`).

test/HashTreeBridgeTest.ts
```diff
@@ -0,0 +1,334 @@
...
+            expect(await hashTreebridgeContract.depositNonce()).to.equal(data1.nonce);
+            expect(await hashTreebridgeContract.depositRoot()).to.equal(new_root);
+
+            expect(await hashTreebridgeContract.claimableTo(data1.nonce)).to.equal(data1.to);
+            expect(await hashTreebridgeContract.claimableAmount(data1.nonce)).to.equal(data1.amount);
+            await expect(tx).to.emit(hashTreebridgeContract, "Claimable").withArgs(data1.nonce, data1.amount, data1.to);
+
+            await fundContract(hashTreebridgeContract, relayer);
+            await expect(hashTreebridgeContract.connect(validator1).claim(data1.nonce)).to.be.revertedWith("Transfer failed");
+        });
```
mialbu: Can you add a case where in the recipient contract is payable?

mialbu: Also, just for the sake of it, can you add a check in the end of this test, to verify that the claimable mapping still contains the nonce, and that the recipient contract and the bridge contracts (well here they are the same, so only one required :)) have the same balance after the tx.

mialbu (changes requested):

test/HashTreeBridgeTest.ts
cloud8little: > Can you add a case where in the recipient contract is payable?

so a new test contract will be added to the contract directory, is that ok?

mialbu: Yes, I think it's fine - just please make sure it's clear that this contract is used for tests, e.g., include `...Test...` in its name.

PR #25, opened 2023-11-15
add zero address check

cloud8little:
add zero address in withdraw and deposit function.

### DIFF
--- a/contracts/AppLogsBridgeContract.sol
+++ b/contracts/AppLogsBridgeContract.sol
@@ -128,7 +128,6 @@ contract AppLogsBridgeContract {
             if (!verify(_depositRoot, depositEntry)) {
                 revert("Invalid proof provided for a deposit.");
             }
-            require(depositEntry.to != address(0), "Address must not be the zero address");
             if (!isContract(depositEntry.to)) {
                 uint256 sendValue = addTenDecimals(depositEntry.amount);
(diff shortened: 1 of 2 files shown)


txhsl (approved):

contracts/AppLogsBridgeContract.sol
```diff
@@ -128,6 +128,7 @@ contract AppLogsBridgeContract {
             if (!verify(_depositRoot, depositEntry)) {
                 revert("Invalid proof provided for a deposit.");
             }
+            require(depositEntry.to != address(0), "Address must not be the zero address");
```
mialbu: Please remove this here, it could potentially break the whole bridge. It's enough if we do the check on the receiving contract. We shouldn't revert based on this on the destination chain.

contracts/HashTreeBridgeContract.sol
```diff
@@ -113,6 +113,7 @@ contract HashTreeBridgeContract {
         uint depositsLength = _deposits.length;
         for (uint i = 0; i < depositsLength; i++) {
             DepositData calldata depositData = _deposits[i];
+            require(depositData.to != address(0), "Address must not be the zero address");
```
mialbu: Remove this as well.

mialbu (changes requested):

PR #30, opened 2023-11-30
add withdrawHash and withdrawRoot to the withdraw event

cloud8little:


mialbu (approved):
Great!

@cloud8little 
Just fyi, if you change a contract, please run the following two commands afterwards (and commit the changes preferrably in a different commit than the contract changes):

Recompile the artifacts of the contracts that we use in the tests (i.e., the files in the `typechain-types` directory):
```bash
npx hardhat compile
```

Run the tests and see whether a test fails now and has to be updated. Usually with contract changes, a test will (or should) fail.
```bash
npx hardhat test
```

PR #43, opened 2024-02-21
Implement Role management

### LINKED ISSUE #35: Implement role management
This issue inolves implementing a separate native Contract `BridgeManagement` replicating the the approach of the `BridgeManagement` contract of the [Neo N3 side](https://github.com/bane-labs/bridge-neo-contracts/blob/70354a6459f5117ece9e9110a1878a45a1d41c82/src/main/java/network/bane/BridgeManagementContract.java).

cloud8little:
close #35


contracts/BridgeContract.sol
```diff
@@ -201,7 +187,7 @@ contract BridgeContract {
         uint j;
         for (uint i = 0; i < 5; i++) {
             for (j = n; j < 7; j++) {
-                if (recovered[i] == validators[j]) {
+                if (recovered[i] == managementContract.validators(j)) {
```
mialbu: We might want to introduce a getter method in the Management contract to get all validators. It's less expensive to call the management contract once getting all validator addresses and then iterate through them locally, instead of calling the management contract for every address.

I've done a small example, and using a method to fetch all validators once and then read them locally is indeed consuming less gas.

mialbu: On the other hand I know this is something that will likely lead to a flag in a future audit, since we're not restricting the size of the validator array in the management contract. So, the contracts are built in a way that it could happen that many validators are set with a considerably lower threshold than the number of validators and then this call would become quite large that could lead to some problems.

My experience with the EVM in this regard is not that deep, maybe someone else has some knowledge of what's best practices on handling such arrays (maybe @chenquanyu)?

contracts/BridgeContract.sol
```diff
@@ -201,7 +187,7 @@ contract BridgeContract {
         uint j;
         for (uint i = 0; i < 5; i++) {
             for (j = n; j < 7; j++) {
```
mialbu: Now that the validators can be updated, and with it the number of validators and the threshold, these numbers can't stay magic numbers and need to be set accordingly based on the validator array in the management contract.

mialbu: Same applies to the lines 179/180:

```solidity
        address[] memory recovered = new address[](5);
        for (uint i = 0; i < 5; i++) {
```

mialbu (changes requested):
I have thought about the design of the Management contract quite a bit now. Note, that I've written the comments on this review two days ago and I'll just leave them for more context if needed. But please mainly act based on this comment here:

With the prospect of maybe having more bridge contracts (e.g. non-/fungible tokens, arbitrary messages), there likely will be overlapping parts. These parts will likely include the same role assignment. However, some other parts, such as a minimum deposit or the size of the fee are more likely to differ. Hence, I'd like to use the management contract to centralize the roles and keep the parameters in the bridge contract.

This means, that just like in the Neo N3 bridge contracts' design, the parameters should be handled in the bridge contract directly. This includes:
- `withdrawalFee`
- `minWithdrawalAmount`
- `maxWithdrawalAmount`
- `maxDepositsPerDistribution`

The management contract then will hold all roles, i.e., all addresses. This includes:
- `owner`
- `relayer`
- `validators`
- `governor`
- `securityGuard`

Unless this should change based on a future design decision, I'd like to keep this design the same as in the current Neo N3 implementation.

In a further step **(not to be included in this PR!)**, we could argue that the `locked` parameter should also be kept in the management contract. If in the future more contracts would depend on the management contract, this would simplify locking the bridge.

So, in conclusion, please change the PR according to the above mentioned design.

contracts/BridgeManagementContract.sol
```diff
@@ -0,0 +1,123 @@
...
+    event SetOwner(address owner);
+    event SetRelayer(address relayer);
+    event SetValidators(address[] validators, uint threshold);
+    event SetGovernor(address governor);
+    event SetSecurityGuard(address securityGuard);
+
+    event WithdrawalFeeChanged(uint256 newFee);
+    event MinWithdrawalAmountChanged(uint256 newAmount);
+    event MaxWithdrawalAmountChanged(uint256 amount);
+    event MaxDepositsPerDistributionChanged(uint8 amount);
```
mialbu: Delete these, they're used only in the BridgeContract.

contracts/BridgeManagementContract.sol
```diff
@@ -0,0 +1,123 @@
...
+    function hasDuplicate(address[] calldata addresses) private returns (bool) {
+        for (uint i = 0; i < addresses.length; i++) {
+            if (addressExists[addresses[i]]) {
+                return true;
+            } else {
+                addressExists[addresses[i]] = true;
+            }
+        }
+        return false;
+    }
```
mialbu: Is this the recommended way of checking for duplicates? No objection about the implementation, I'm just genuinely interested.

contracts/BridgeManagementContract.sol
```diff
@@ -0,0 +1,123 @@
...
+
+    function setValidators(
+        address[] calldata _validators,
+        uint threshold
+    ) external onlyOwner {
+        require(
+            _validators.length > 0,
```
mialbu: `_validators.length` is used several times in the scope of the `setValidators` method. Please assign it to a local variable and then use that. This will reduce the gas cost of this method.

contracts/BridgeContract.sol
```diff
@@ -190,17 +182,20 @@ contract BridgeContract {
...
         }
         // check if all recovered addresses are in the validator set
         uint covered = 0;
         uint n = 0;
         uint j;
-        for (uint i = 0; i < 5; i++) {
-            for (j = n; j < 7; j++) {
+        address[] memory validators = managementContract.getValidators();
+        for (uint i = 0; i < threshold; i++) {
+            for (j = n; j < validators.length; j++) {
```
mialbu, line 196: Use a temporary variable for `validator.length` to lower the gas cost.

```suggestion
        address[] memory validators = managementContract.getValidators();
        uint256 nrValidators = validators.length;
        for (uint i = 0; i < threshold; i++) {
            for (j = n; j < nrValidators; j++) {
```

contracts/BridgeContract.sol
```diff
@@ -181,7 +172,8 @@ contract BridgeContract {
         Signature[] calldata _signatures
     ) private view returns (bool) {
         require(
-            _signatures.length == requiredValidatorSignaturesForDeposit,
+            _signatures.length ==
+                managementContract.requiredValidatorSignaturesForDeposit(),
```
mialbu: Use a temporary variable for `managementContract.requiredVAlidatorSignaturesForDeposit()`, e.g., `valdiatorThreshold`. It's reused on the lines 185/6, which will become obsolete, so you can remove the lines 185/6 after this change.

test/BridgeTest.ts
```diff
@@ -42,22 +46,65 @@ describe("Bridge contract", function () {
...
+        it("Set invalid min and max withdrawal amounts", async function () {
+            const { bridgeContract: bridgeContract, governor } = await loadFixture(deployBridgeFixture);
+            const minWithdrawalAmount = await bridgeContract.minWithdrawalAmount();
+            const maxWithdrawalAmount = await bridgeContract.maxWithdrawalAmount();
+
+            let tx = bridgeContract.connect(governor).setMinWithdrawalAmount(maxWithdrawalAmount);
+            await expect(tx).to.be.revertedWith("Amount must be less than the maximal withdrawal amount");
+
+            tx = bridgeContract.connect(governor).setMaxWithdrawalAmount(minWithdrawalAmount);
+            await expect(tx).to.be.revertedWith("Amount must be greater than the minimal withdrawal amount");
```
mialbu: Add two more cases, one for `setMinWithdrawalAmount` with an amount lower than the current min, and one for the `setMaxWithdrawalAmount` with an amount higher than the current max.

test/BridgeManagementTest.ts
mialbu, line 1: Can you use a different account for the owner than the validator1. It can be confusing when reading the tests, e.g., the test `Set owner` in the `Bridge role setting` scope.

mialbu (changes requested):
Great, we're almost there. I have some comments that should be addressed before we merge.

contracts/BridgeManagementContract.sol
cloud8little: > Is this the recommended way of checking for duplicates? No objection about the implementation, I'm just genuinely interested.

I think map is usally used for this, only iterate once, I am also thinking about other efficient ways to do this, let me know your thoughts.

test/BridgeTest.ts
cloud8little: > Add two more cases, one for `setMinWithdrawalAmount` with an amount lower than the current min, and one for the `setMaxWithdrawalAmount` with an amount higher than the current max.

it is in the scope of valid cases.  `setMinWithdrawalAmount` with an amount lower than the current min is the case in Line 64, I've added one more case for `setMaxWithdrawalAmount` with an amount higher than the current max.

mialbu: Oh, I wrote incorrectly. I meant invalid cases:

- `setMaxWithdrawalAmount` with an amount lower than the current min

and

- `setMinWithdrawalAmount` with an amount higher than the current max

contracts/BridgeManagementContract.sol
mialbu: It works, but only if after a first change, the next one mustn't include any address that's already a validator.

So, let's assume the validators are changed for the first time with any `n` validators. After that change, if one validator address needs to be updated, including any of the other `n-1` existing validators will make the transaction fail. So, we'd need to delete the complete mapping before returning `false` in this function, or we can't reuse any existing address when we need to update them.

mialbu: I've rewritten it to a simple iteration, since the validator array will not be large anyway. Also, this way we don't have to deal with the above mentioned case.

Check commit (3aaa7d7). It contains a test case that should pass (just replace one validator address), but doesn't with using a mapping in storage for the duplicate check. When updating the `hasDuplicate` function to a simple iteration sequence in commit (2871bbe), this test case is successful.

test/BridgeTest.ts
mialbu: Done in 0edf094

mialbu (approved):

PR #44, opened 2024-03-26
Allow Bridge Upgrade

### LINKED ISSUE #38: Contract Update

mialbu:
Resolves #38


contracts/BridgeImpl.sol
```diff
@@ -1,33 +1,15 @@
 // SPDX-License-Identifier: MIT
-pragma solidity ^0.8.9;
+pragma solidity ^0.8.20;
 
 import "./BridgeManagementContract.sol";
+import "./BridgeStorage.sol";
+import "@openzeppelin/contracts-upgradeable/proxy/utils/Initializable.sol";
 
 // Todo: Before compiling byte code for genesis script, make sure to remove the receive function as it is only used for testing purpose.
```
gsmachado: is this TODO still valid?

contracts/BridgeProxy.sol
```diff
@@ -0,0 +1,52 @@
...
+            case 0 {
+                revert(ptr, size)
+            }
+            default {
+                return(ptr, size)
+            }
+        }
+    }
+
+    receive() external payable {
```
gsmachado: If I'm not mistaken, you can use a feature in hardhat (hardhat's built-in network) that allows for direct manipulation and interaction with the blockchain state (to set-up some test scenarios)... 

Then, maybe can we remove this `receive()` from the contract? Or am I missing something here?

Adding code just required for tests to the smart contract is a dangerous practice, after all. 😅

contracts/BridgeImpl.sol
```diff
@@ -1,33 +1,15 @@
...
-    uint256 public withdrawalFee = 10000000_0000000000;
-    uint256 public minWithdrawalAmount = 1_00000000_0000000000;
-    uint256 public maxWithdrawalAmount = 10000_00000000_0000000000;
-    uint8 public maxDepositsPerDistribution = 100;
-
-    // Events
-
+contract BridgeImpl is BridgeStorage, Initializable {
+    event Locked();
+    event Unlocked();
```
gsmachado, line 10: should we add any additional parameters to the locked and unlocked events? like, block number, timestamp, etc? or do you think it's not necessary?

gsmachado: hmmm, IMHO, I'd rename this to `Lock` and `Unlock`.

gsmachado (commented):
Some comments to be considered.

Moreover: I believe it's not a good practice to commit `typechain-types`. It'd be good to check what other projects do.

contracts/BridgeProxy.sol
chenquanyu: We can just use openzeppelin proxy like governance proxy https://github.com/bane-labs/go-ethereum/blob/governance/contracts/solidity/TransparentUpgradeableProxy.sol

contracts/BridgeImpl.sol
mialbu: yes

contracts/BridgeImpl.sol
mialbu: > should we add any additional parameters to the locked and unlocked events? like, block number, timestamp, etc? or do you think it's not necessary?

We can. Can you open an issue for this?



> hmmm, IMHO, I'd rename this to `Lock` and `Unlock`.

This is a design choice, like other events are called `...Changed` and not `...Change`. The only deviant events are `Deposit` and `Withdrawal`.

contracts/BridgeProxy.sol
mialbu: > We can just use openzeppelin proxy like governance proxy https://github.com/bane-labs/go-ethereum/blob/governance/contracts/solidity/TransparentUpgradeableProxy.sol

We can for T3. Although, I dislike the approach for future iterations because it restricts the proxy to have a single-sig owner. Singe having a multi-sig owner is not in scope for T3, we'll need to go to T4 anyways to change that.

Another issue I have with this implementation is its restriction of not letting the admin fallback to the implementation contract. While this is not a problem for the bridge contract, it'll be one for the management contract. So, once we make the management contract, we'd need to either:
- use a different "owner" on the management contract than the proxy's `admin`, i.e., distinguish the responsibilities of upgrading a contract and assigning roles to two different roles.
- use a different proxy contract than the `TransparentUpgradeableProxy` that allows the `owner` to both upgrade the implementation and assign roles.

mialbu: > Adding code just required for tests to the smart contract is a dangerous practice, after all. 😅

I agree.

> If I'm not mistaken, you can use a feature in hardhat (hardhat's built-in network) that allows for direct manipulation and interaction with the blockchain state (to set-up some test scenarios)...

Do you know what that feature is called?

chenquanyu: >We can for T3. Although, I dislike the approach for future iterations because it restricts the proxy to have a single-sig owner. Singe having a multi-sig owner is not in scope for T3, we'll need to go to T4 anyways to change that.

We now have the proxy admin contract like multi-sig, please refer to https://github.com/bane-labs/go-ethereum/blob/governance/contracts/solidity/GovProxyAdmin.sol

>Another issue I have with this implementation is its restriction of not letting the admin fallback to the implementation contract. 

This is quite an issue if you want call the implementation contract from proxy, but this is for security reason. Please refer to https://ethereum.stackexchange.com/questions/123343/why-is-admin-restricted-from-accessing-fallback-function-in-transparent-proxy

mialbu: > If I'm not mistaken, you can use a feature in hardhat (hardhat's built-in network) that allows for direct manipulation and interaction with the blockchain state (to set-up some test scenarios)...
> 
> Then, maybe can we remove this `receive()` from the contract? Or am I missing something here?
> 
> Adding code just required for tests to the smart contract is a dangerous practice, after all. 😅

I agree 100%. I've actually found a way now. In the tests the contract's address is precomputed based on the deployers address and nonce. That address is funded and **only then** the contract is deployed, leading to the now deployed contract having a balance.

contracts/BridgeImpl.sol
mialbu: I've opened an issue for this: #47.

chenquanyu (approved):

PR #45, opened 2024-03-27
UUPS Upgrade Pattern

mialbu:
@chenquanyu Please take a look at this PR. It's not complete yet though. For example, we should discuss the `endUpgradeability` feature and if we'd like to have it, write a test for it. No worries, I can still remove it for this version if it's not desired.

Even if we don't add it in the beginning, we could still add it in an upgraded version later as well. I added it initially because I've followed the discussions in Discord about removing upgradeability in bNeo and it may be a good feature to have for the long term for the bridge as well.

There are some advantages I see in using the UUPS upgrade pattern over the transparent proxy pattern:
- There'll be the option of removing upgradeability without implementing the functionality in the beginning (*1)
- It's more flexible with using the management's roles.

So, generally UUPS seems more flexible, since it allows for a more specific change in the upgrade options, while the transparent upgrade pattern is restricted in the un-upgradeable proxy.

*1 in the transparent pattern, the functionality of restricting an upgrade must be implemented in the proxy itself. Thus, it must be there in the very beginning of Neo X, while with UUPS, the contract can be initially deployed without the feature, but it can be added in a future upgrade version.


contracts/BridgeStorage.sol
```diff
@@ -46,6 +46,40 @@ contract BridgeStorage is Initializable {
         });
     }
 
+    function _authorizeUpgrade(
+        address newImplementation
+    ) internal virtual override onlyOwner {}
```
chenquanyu: Perhaps it would be better to also use `GovProxyAdmin` contract to control genesis initialized contracts upgrade. Please refer to https://github.com/bane-labs/go-ethereum/pull/7/files.

mialbu: Adapted for both bridge and management contract.

PR #75, opened 2024-04-16
T3

mialbu:


release/testnet/T3/v1/Bridge/proxy-storage.json
```diff
@@ -0,0 +1,50 @@
...
+        "key": "0x360894a13ba1a3210667c828492db98dca3e2076cc3735a920a3ca505d382bbc",
+        "value": "0x0000000000000000000000001212100000000000000000000000000000000004"
+    },
+    "0x290decd9548b62a8d60345a988386fc84ba6bc95484008f6362f93160ef3e563": {
+        "key": "0x0000000000000000000000000000000000000000000000000000000000000000",
+        "value": "0x0000000000000000000000001212000000000000000000000000000000000005"
+    },
+    "0xb10e2d527612073b26eecdfd717e6a320cf44b4afac2b0732d9fcbe2b7fa0cf6": {
+        "key": "0x0000000000000000000000000000000000000000000000000000000000000001",
+        "value": "0x0000000000000000000000000000000000000000000000000000000000000000"
```
AnnaShaleva: What do you think about removal storage items with empty values? Solidity doesn't distinguish empty value from missing storage item, IIUC.

AnnaShaleva: Although it needs to be tested.

AnnaShaleva (commented):
The compiled binaries and storage part LGTM, ported to https://github.com/bane-labs/go-ethereum/blob/v0.1.0/config/genesis_testnet.json.

release/testnet/T3/v1/Bridge/proxy-storage.json
mialbu: I think you're right, these are not needed actually. Every value that is accessed in storage and nothing is stored there it'll just be zeros by default. Also leading zeros are not needed as well. I'll change it.

PR #85, opened 2024-05-02
Fungible Token Integration

### LINKED ISSUE #55: NBE-08 - Solidity literals `[Informative]`
<p>In order to make the code easier to read and to minimize human errors, Solidity recommends the use of literals which consequently makes it more user friendly.</p><p>In the case of <code>BridgeContract</code>, the <code>10**10</code> operation is frequently used to add or eliminate the decimal difference between gas and ether. This multiplication can be avoided by using the literal <code>1e10</code>.</p>

image

**References:**
- https://docs.soliditylang.org/en/latest/units-and-global-variables.html#units-and-globally-available-variables

**Source references:**
<p><a target="_blank" rel="noopener noreferrer nofollow" href="https://github.com/bane-labs/bridge-evm-contracts/blob/14cd3efd3f618fd18a4cda358c791fdca0e66cff/contracts/BridgeContract.sol#L335">https://github.com/bane-labs/bridge-evm-contracts/blob/14cd3efd3f618fd18a4cda358c791fdca0e66cff/contracts/BridgeContract.sol#L335</a></p>
<p><a target="_blank" rel="noopener noreferrer nofollow" href="https://github.com/bane-labs/bridge-evm-contracts/blob/14cd3efd3f618fd18a4cda358c791fdca0e66cff/contracts/BridgeContract.sol#L330">https://github.com/bane-labs/bridge-evm-contracts/blob/14cd3efd3f618fd18a4cda358c791fdca0e66cff/contracts/BridgeContract.sol#L330</a></p>
<p><a target="_blank" rel="noopener noreferrer nofollow" href="https://github.com/bane-labs/bridge-evm-contracts/blob/14cd3efd3f618fd18a4cda358c791fdca0e66cff/contracts/BridgeContract.sol#L276">https://github.com/bane-labs/bridge-evm-contracts/blob/14cd3efd3f618fd18a4cda358c791fdca0e66cff/contracts/BridgeContract.sol#L276</a></p>

### LINKED ISSUE #83: Add support for ERC20

### LINKED ISSUE #84: Use `uint256` for `nonce` and `amount`

### LINKED ISSUE #86: Use `ReentrancyGuard`'s `nonReentrant` modifier for external contract calls

### LINKED ISSUE #87: Use Contract Addresses as Identifiers in Token Bridges
For the token bridges, we're currently using uint256 as identifier. However, instead of introducing an additional value, it would be better if we'd use already existing identifier values, i.e., the respective contract addresses on either chain. Instead of prepending an identifier to a deposit/withdrawal hash, we should prepend the N3 token address and the Neo X token address.

### LINKED ISSUE #88: Support NEO

mialbu:
This PR is still WIP.

Closes #86 
Closes #84 
Closes #83 
Closes #55
Closes #87
Closes #88

- Added `nonReentrant` modifier to `deposit()`, `claim()`, `depositToken()` and `claimToken()`


contracts/interfaces/IERC20Capped.sol
```diff
@@ -0,0 +1,9 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.24;
+
+interface IERC20Capped {
```
txhsl: `ERC20Capped` is an existing template in [OpenZeppelin](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/release-v5.0/contracts/token/ERC20/extensions/ERC20Capped.sol), I think they are different things, maybe rename this interface.

If this interface is meant to allow ERC-20 transfer, then just import https://github.com/OpenZeppelin/openzeppelin-contracts/blob/release-v5.0/contracts/token/ERC20/IERC20.sol, instead of creating a new one.

contracts/library/BridgeLib.sol
```diff
@@ -67,25 +67,13 @@ library BridgeLib {
     }
 
     // Removes 10 decimal points from the amount. GasToken originally has 8 decimals and on this chain it has 18 decimals.
-    function _removeTenDecimals(uint256 _value) internal pure returns (uint64) {
-        return uint64(_value / (10 ** 10));
+    function _removeTenDecimals(
+        uint256 _value
+    ) internal pure returns (uint256) {
+        return uint256(_value / (10 ** 10));
```
txhsl: Maybe we don't need this `uint256()` convert.

contracts/base/IGasBridge.sol
```diff
@@ -0,0 +1,43 @@
...
+        bytes32 withdrawalRoot
+    );
+    event WithdrawalFeeChange(uint256 newFee);
+    event MinWithdrawalAmountChange(uint256 newAmount);
+    event MaxWithdrawalAmountChange(uint256 amount);
+    event MaxDepositsPerDistributionChange(uint8 amount);
+
+    // Gas-related functions
+
+    function deposit(
```
txhsl: `depositGAS`.

contracts/base/ITokenBridge.sol
```diff
@@ -0,0 +1,75 @@
...
+    function unlockToken(uint256 id) external;
+
+    function setTokenWithdrawalMinAmount(uint id, uint256 minAmount) external;
+
+    function setTokenWithdrawalMaxAmount(
+        uint256 id,
+        uint256 maxAmount
+    ) external;
+
+    function depositToken(
```
txhsl: If we are going to allow ERC20 here, then use `depositERC20` and `ERC20Bridge`.

contracts/base/IBridge.sol
```diff
@@ -0,0 +1,18 @@
...
+interface IBridge {
+    // General bridge events
+
+    event Lock();
+    event Unlock();
+    event Fund(uint256 amount);
+
+    // General bridge functions
+
+    function lock() external;
```
txhsl: Maybe move this `lock()` and `unlock()` to `IGasBridge.sol`.

contracts/library/BridgeStorageTypes.sol
```diff
@@ -0,0 +1,64 @@
...
+        uint256 fee;
+        uint256 minAmount;
+        uint256 maxAmount;
+        uint8 maxDepositsPerDistribution;
+        uint256[2] gap; // not needed
+    }
+
+    // Token Bridges
+
+    enum TokenType {
```
txhsl: Are we going to support `ERC20Uncapped`? I'm afraid not, and I don't think it will be compatible with method `depositToken()`.

For example, `ERC721` bridge is somehow different from `ERC20` Bridge, there will be a `tokenId` in transfer method, instead of the `value`. Please refer to https://github.com/OpenZeppelin/openzeppelin-contracts/blob/release-v5.0/contracts/token/ERC721/IERC721.sol. We may need another bridge lib for this.

contracts/base/BridgeImpl.sol
```diff
@@ -0,0 +1,392 @@
...
+                        to
+                    );
+                    emit Claimable(depositEntry.nonce, depositEntry.amount, to);
+                }
+            }
+        }
+    }
+
+    // Anyone can execute a claim. The funds of a claimable will be sent to the defined address in the claimableTo mapping.
+    function claim(uint256 _nonce) external unlocked {
```
txhsl: It is necessary to use `ReentrancyGuard` when there is a message calling to outside (e.g. `to.call{}()`), to prevent potential attack, for the safety of future maintenance. Please refer to https://github.com/OpenZeppelin/openzeppelin-contracts/blob/release-v5.0/contracts/utils/ReentrancyGuard.sol.

contracts/base/BridgeImpl.sol
```diff
@@ -0,0 +1,392 @@
...
+        return _tokenContract.transfer(_to, _amount);
+    }
+
+    /**
+     * @notice Withdraw tokens to Neo N3.
+     * @dev This function must be called by the contract that is registered and thus is responsible for bridging the tokens. This can but need not be the token contract itself.
+     * @param _amount the amount of tokens to withdraw.
+     * @param _to the address to which the tokens should be sent.
+     */
+    function withdrawToken(uint256 _amount, address _to) external override {
```
txhsl: The fee logic is missing here.

contracts/base/BridgeStorage.sol
```diff
@@ -0,0 +1,372 @@
...
+    IBridgeManagement public management;
+    bool public locked;
+    // Gas Bridge
+    BridgeStorageTypes.GasBridge public gasBridge;
+    mapping(uint256 => BridgeStorageTypes.Claimable) public claimableGas;
+    // Token Bridges
+    mapping(BridgeStorageTypes.TokenType tokenType => BridgeStorageTypes.TokenTypeConfig)
+        public tokenTypeConfigs;
+    mapping(address tokenAddress => uint256 id) public tokenIds;
+    mapping(uint256 id => BridgeStorageTypes.TokenBridge) public tokens;
```
txhsl: Maybe rename this `tokens` to `bridges` or something else.

contracts/base/BridgeStorage.sol
```diff
@@ -0,0 +1,372 @@
...
+    mapping(uint256 id => mapping(uint256 nonce => BridgeStorageTypes.Claimable))
+        public tokenClaimables;
+
+    // End Storage Slots
+
+    constructor(address _management) {
```
txhsl: The initialization of `TokenTypeConfig` is necessary.

contracts/library/BridgeStorageTypes.sol
```diff
@@ -0,0 +1,64 @@
...
+        // ERC721
+    }
+
+    struct TokenTypeConfig {
+        uint256 fee;
+        uint8 maxDepositsPerDistribution;
+    }
+
+    struct TokenBridge {
+        bool registered; // used for simple existence check
```
txhsl: The storage of `registered` is a bit of redundant. It is not expensive, but maybe check something else.

contracts/base/BridgeStorage.sol
```diff
@@ -0,0 +1,372 @@
...
+        if (_id == 0) revert InvalidTokenId();
+        // Check if token config contains valid values
+        if (_tokenConfig.minAmount > _tokenConfig.maxAmount)
+            revert InvalidAmount();
+        if (_tokenConfig.contractAddress == address(0)) revert InvalidAddress();
+
+        // Check if token bridge is already registered
+        BridgeStorageTypes.TokenBridge memory tokenBridge = tokens[_id];
+        if (tokenBridge.registered) revert TokenAlreadyRegistered(_id);
+        assert(tokenIds[_tokenConfig.contractAddress] == 0);
```
txhsl: The `id` and `address` are both unique identifiers. Why not just use the contract address as the key of the mapping, so that we don't have to bind it with its `_id` and make this check?

contracts/base/BridgeStorage.sol
```diff
@@ -0,0 +1,372 @@
...
+
+    function _setTokenMaxWithdrawalAmount(
+        uint256 _id,
+        uint256 _amount
+    ) internal {
+        if (_amount < tokens[_id].config.minAmount) revert InvalidAmount();
+        tokens[_id].config.maxAmount = _amount;
+    }
+
+    function _getTokenTypeConfig(
```
txhsl: The setter for `TokenTypeConfig` seems missing.

txhsl (changes requested):

txhsl:
Sorry, I know this PR is WIP, but I'm requested to have a review. :)

contracts/base/ITokenBridge.sol
mialbu: You're right, I'll rename it. However, to give you some context: I've called it `depositToken` because with the current design around creating new hashes from token deposits, it will be very simple to include ERC721 tokens as well, as long as only the `tokenId` is needed and no metadata needs to be bridged. For example, an NFT mint or transfer also requires an `address` and a `uint256` parameter which suits the current design around creating a hash from that data.

👇🏼 

In order to restrict any signature replay attacks, I've added an identifier to the hash process for tokens. For GAS it's `sha256(concat(nonce, amount, to))` (i.e., `sha256(concat(uint256, uint256, address))`), and for tokens it's `sha256(concat(id, nonce, amount, to))` (i.e., `sha256(concat(uint256, uint256, uint256, address))`), where the `amount` could act as the `tokenId` for NFTs.

contracts/base/BridgeImpl.sol
mialbu: Done

contracts/base/BridgeStorage.sol
mialbu: > The `id` and `address` are both unique identifiers. Why not just use the contract address as the key of the mapping, so that we don't have to bind it with its `_id` and make this check?

They are, but only on either chain. The idea behind using a different unique identifier than the ERC20 address is that it can be used on N3 as well. Imo, we can use the ERC20 address as unique identifier, but then we'd need to store it's "counterpart" address of the N3 NEP-17 token here as well. I'd like that it's clear which hash chain belongs to which hash chain between Neo X and Neo N3, and this is achieved with an id in a nicer and cleaner way imho.

contracts/library/BridgeLib.sol
mialbu: Addressed in [746f151](https://github.com/bane-labs/bridge-evm-contracts/pull/85/commits/746f15126ef3a85b4b16703eb49b41192591af64)

contracts/interfaces/IERC20Capped.sol
mialbu: Already did in the meantime 👍🏼

contracts/library/BridgeStorageTypes.sol
mialbu: We can support `ERC20Uncapped`, just as well as `ERC721`. We just need to add some specific functionality instead of just calling `ERC20.transfer()`. The idea of the current `depositToken` function implementation is that based on the token type, different logic is executed on the destination contract, which might be a `ERC20Capped`, and `ERC20Uncapped` or an `ERC721`. There are some restrictions, but it's absolutely possible to do it. If a token owner doesn't want to give ownership to the bridge contract (i.e., to mint), we can introduce an intermediary contract, that will be called from the bridge contract, and that contract will be the token contract's owner.

contracts/library/BridgeStorageTypes.sol
mialbu: Fixed in [89a07ab](https://github.com/bane-labs/bridge-evm-contracts/pull/85/commits/89a07ab76462b0cb9b3aea1644ad0feb6b708ca1). Now using a check whether the token bridge's config's `contractAddress` is `address(0)`.

contracts/base/BridgeStorage.sol
mialbu: Added in [cb16577](https://github.com/bane-labs/bridge-evm-contracts/pull/85/commits/cb165778a17c41e0902202e10eeba09a364c7d30).

contracts/base/IBridge.sol
mialbu: Depends whether we want to only lock the gas bridge separately. For tokens I think it's nice that they can be locked and unlocked individually. For Gas, I'm not so sure. Either way, I think we should have a complete lockdown functionality, and as it is now, `lock()` and `unlock()` could* act nicely as such. If we want to add functionality to lock the gas bridge individually, I think it's better to introduce a new value in storage with new functions, e.g., `un/lockGasBridge()`.

\*They currently don't, but it would suit the current storage design that we just add the locked/unlocked modifiers to the token functions as well.

contracts/base/BridgeImpl.sol
mialbu: Addressed in [f7171b7](https://github.com/bane-labs/bridge-evm-contracts/pull/85/commits/f7171b759d5f09769d1a3f55b0041f1732bf9b8e), please take a look.

contracts/base/BridgeStorage.sol
mialbu: Renamed to `tokenBridges` and `tokenBridgeIds`.

contracts/base/BridgeStorage.sol
mialbu: With the newly added function `setTokenTypeConfig()` it's not mandatory to have in the constructor. Also, once we use this contract for upgrading in T3, we need to use this setter function anyway. Once we go to mainnet, we should precompute its storage entries, or we don't and just need to be aware that we need to invoke the setter before registering any tokens of unset tokenTypeConfigs.

contracts/base/IGasBridge.sol
mialbu: I'll leave that open for now, but I tend to agree to a renaming here, maybe `depositGas` works as well.

mialbu: I've renamed gas-related exported functions in [51a969d](https://github.com/bane-labs/bridge-evm-contracts/pull/85/commits/51a969d17b6831246f2ac280f75c81fb0ea6c860), i.e., `depositGas()`, `claimGas()` and `withdrawGas()`. If you agree with the change, please mark this convo as resolved.

contracts/base/BridgeStorage.sol
txhsl: Fine.

contracts/base/BridgeStorage.sol
txhsl: > be aware that we need to invoke the setter before registering any tokens of unset tokenTypeConfigs.

Yes, this is important.

contracts/base/IBridge.sol
txhsl: AFAIK, the `unlocked` modifier is only affecting `depositGas`, `claimGas` and `withdrawGas`, so it seems like a gas-bridge-only lock. Just put it at the correct place when we finalize the contract.

contracts/base/BridgeImpl.sol
txhsl: Should we charge GAS or the bridged token?

contracts/library/BridgeStorageTypes.sol
txhsl: 🤔Is there any case that we may bridge `ERC20Uncapped` on N3 or Neo X?

contracts/base/ITokenBridge.sol
txhsl: Well, but take care because the `uint256` in `Claimable` data have different meanings.

contracts/library/BridgeStorageTypes.sol
mialbu: > Is there any case that we may bridge `ERC20Uncapped` on N3 or Neo X?

Depends on demand ofc, but e.g., FLM has dynamic total supply afaik.

contracts/base/BridgeImpl.sol
mialbu: > Should we charge GAS or the bridged token?

Good question. Although, would we really want a token? @steven1227

contracts/library/BridgeStorageTypes.sol
txhsl: Dynamic total supply is allowed by both ERC20 and NEP17.

txhsl: Well, if we pack ERC20 and ERC721 in the same data structure, then we can leave this third type TBD.

mialbu: Sure, the idea with the approach of having ERC20Capped and ERC20Uncapped is that on the destination chain when distributing we can differentiate between using `transfer()` and `mint()`.

It depends how we want anyone that requests to be registered should be included in the bridge. What should they provide once we accept a token to be supported by the bridge?

If it's a capped token, they can allocate the total supply on the destination chain to the bridge. Then we can just use `transfer` on the new ERC20 they deploy.

If it's an uncapped token, they might not like to put a hardcap allocation on the destination chain to the bridge, but rather ask if we can somehow mint. For this we either:
- need to support calling a `mint()` function that is `accessManaged`.
- they need to write some logic into their `transfer` function, such that if the `msg.sender` is the bridge, they mint instead of transfer.

Or we just create a well-defined interface that the bridge supports, and require anyone that wants to get registered to deploy a contract that adheres to that interface (and then they can add their logic in there - whether it'll be a mint/transfer/...).

txhsl: > If it's a capped token, they can allocate the total supply on the destination chain to the bridge.

Well, I see the problem.

contracts/base/BridgeImpl.sol
gsmachado: In my opinion, we should charge $GAS.

contracts/library/BridgeStorageTypes.sol
gsmachado: > Well, I see the problem.

Usually, there are more problems than we imagine. 🤷‍♂️

contracts/base/ITokenBridge.sol
gsmachado: @mialbu can we resolve this comment?

mialbu: Yes, I'm well aware of the different meanings. I'll keep the current naming. Once introducing support for NFTs is on the table, we should rename `amount`, e.g., to `value`. Until then, the current state of it is fine.

contracts/library/TokenBridgeLib.sol
```diff
@@ -0,0 +1,91 @@
...
+     * @param _amount The amount to transfer.
+     * @param _to The address of the recipient.
+     */
+    function _executeERC20Transfer(
+        address _neoXToken,
+        uint256 _amount,
+        address _to
+    ) internal returns (bool) {
+        IERC20 neoXToken = IERC20(_neoXToken);
+        return neoXToken.transfer(_to, _amount);
```
txhsl: Can merge to one line.

contracts/bridge/BridgeImpl.sol
```diff
@@ -0,0 +1,514 @@
...
+        emit TokenUnregister(_neoXToken, _getNeoN3Token(_neoXToken));
+    }
+
+    /**
+     * @notice Pause a token bridge. No deposits, withdrawals, or claims of a token bridge can be made while it is locked.
+     * @param _neoXToken the address of the token on the Neo X network.
+     */
+    function pauseTokenBridge(
+        address _neoXToken
+    ) external override onlyTokenBridgeUnpaused(_neoXToken) onlyGovernor {
```
txhsl: I see the gas bridge pause is `onlySecurityGuard`, but here the token bridge is `onlyGovernor`. What is the difference?

contracts/bridge/BridgeImpl.sol
```diff
@@ -0,0 +1,514 @@
...
+        address _neoXToken,
+        address _to,
+        uint256 _amount
+    )
+        external
+        payable
+        override
+        nonReentrant
+        onlyBridgeUnpaused
+        onlyTokenBridgeUnpaused(msg.sender)
```
txhsl: Using `msg.sender` seems not correct.

contracts/library/StorageTypes.sol
```diff
@@ -0,0 +1,69 @@
...
+
+    struct TokenBridge {
+        bool paused;
+        State depositState;
+        State withdrawalState;
+        TokenConfig config;
+    }
+
+    struct TokenConfig {
+        address neoN3Token;
```
txhsl, line 61: Is this a token address on N3? If it is, then is it compatible to be stored in Solidity `address`?

txhsl (changes requested):

contracts/library/StorageTypes.sol
mialbu: It is. As far as I know, there's no checksum validation for address parameters. Thus, any value that is hexadecimal and 20 bytes long (which applies to any Neo N3 scripthash) is compatible.

Only if you provide a constant or initial value in solidity development, an address needs to be checksummed. Meaning the solidity compiler checks any address at compile time, but in the deployed script there's no such check.

contracts/bridge/BridgeImpl.sol
mialbu: No difference. It should be `onlySecurityGuard`. I'll change it.

mialbu: Done in a13ec41c8080ff9f8c5383adad59ecf064326e8f.

contracts/library/TokenBridgeLib.sol
mialbu: Done in 5a0a52c4b6b1afd72190fae053574b7ad013adcd.

contracts/bridge/BridgeImpl.sol
mialbu: You're correct, I've changed it to `_neoXToken` in 9a206c7762c9c1c40aaaa73e32f74b932b2ae8d0.

mialbu:
@txhsl
Fyi, I've also added `_disableInitializers()` in the constructor of the `BridgeStorage.sol` contract in 00ece04b3bf2459dc77a4406f7cd420a7f9372d8.

txhsl (approved):

PR #104, opened 2024-05-24
Add Hash and Root to Withdrawal Events and add DepositRootUpdate Events

### LINKED ISSUE #102: Add hash and new root to token withdrawal event

### LINKED ISSUE #103: Add event for deposit root update

mialbu:


contracts/bridge/BridgeImpl.sol
```diff
@@ -480,7 +491,16 @@ contract BridgeImpl is
             _neoXToken,
             StorageTypes.State({nonce: newNonce, root: newRoot})
         );
-        emit TokenWithdrawal(_neoXToken, newNonce, _to, tokenValue);
+        emit TokenWithdrawal(
```
txhsl, line 494: Is it necessary to emit so many variables? It costs https://github.com/ethereum/go-ethereum/blob/master/params/protocol_params.go#L40.

https://ethereum.stackexchange.com/questions/106772/how-much-gas-does-it-cost-to-emit-an-event

gsmachado: yes, it adds costs.

However, if such costs are minimal, I'm in favor of emitting more info since it could be necessary for monitoring systems or other applications.

Of course, we could always add this info later by upgrading it. That's why I'm not strongly opinionated here.

txhsl: These data can help a lot in atomicity, but it depends on whether the total transaction fee changes a lot.

If the test result is acceptable, then let's solve this conversation and merge the codes.

txhsl (approved):

contracts/bridge/BridgeImpl.sol
mialbu: I'm also not strongly oppinionated, but I would leave it as it is now. The only parameters that I'd potentially would remove are `from` (here `msg.sender`) and `withdrawalHash`.

The other parameters are necessary, because we want to verify the data and the next root only knowing the previous root and given the data from this event. For this we need the first 5 parameters and the new root in the event.

mialbu: I've opened an issue (#105) to keep this in mind in the future. This discussion is resolved with that, but I'll leave it open for more direct visual context if someone were to go back to this PR.

PR #109, opened 2024-05-29
Update TokenBridgeLib.sol

passer-byzhang:
Use safeTransfer to prevent from directly throwing an error when the transfer transaction fails.
For example, [default ERC20 token contract](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/token/ERC20/ERC20.sol#L104) will revert error instead of return a `false` if the transfer failed.

### DIFF
--- a/contracts/bridge/BridgeImpl.sol
+++ b/contracts/bridge/BridgeImpl.sol
@@ -415,7 +415,8 @@ contract BridgeImpl is
                 tokenType == StorageTypes.TokenType.ERC20Capped
         );
         // Note: For NEO tokens, the transfer value has already been extended with 18 decimals in the deposit function.
-        TokenBridgeLib._executeERC20Transfer(_neoXToken, claimable.amount, to);
+        bool success = TokenBridgeLib._executeERC20Transfer(_neoXToken, claimable.amount, to);
+        if(!success) revert TransferFailed();
     }
 
     function _emitTransferEventOrAddNewTokenClaimable(
--- a/contracts/bridge/BridgeImpl.sol
+++ b/contracts/bridge/BridgeImpl.sol
@@ -416,7 +416,7 @@ contract BridgeImpl is
         );
         // Note: For NEO tokens, the transfer value has already been extended with 18 decimals in the deposit function.
         bool success = TokenBridgeLib._executeERC20Transfer(_neoXToken, claimable.amount, to);
-        if(!success) revert TransferFailed();
+        if (!success) revert TransferFailed();
     }
 
     function _emitTransferEventOrAddNewTokenClaimable(
--- a/contracts/library/TokenBridgeLib.sol
+++ b/contracts/library/TokenBridgeLib.sol
@@ -19,7 +19,7 @@ library TokenBridgeLib {
     ) internal returns (bool) {
         bytes memory transferCall = abi.encodeCall(IERC20.transfer, (_to, _amount));
         (bool success, bytes memory returndata) = address(_neoXToken).call(transferCall);
-        return success && (returndata.length == 0 || abi.decode(returndata, (bool))) && address(_neoXToken).code.length > 0;
+        return success && (returndata.length == 0 || abi.decode(returndata, (bool)));
     }
 
     /**
--- a/contracts/library/TokenBridgeLib.sol
+++ b/contracts/library/TokenBridgeLib.sol
@@ -17,7 +17,9 @@ library TokenBridgeLib {
         uint256 _amount,
         address _to
     ) internal returns (bool) {
-        return IERC20(_neoXToken).transfer(_to, _amount);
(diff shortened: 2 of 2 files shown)


contracts/library/TokenBridgeLib.sol
```diff
@@ -17,7 +17,9 @@ library TokenBridgeLib {
         uint256 _amount,
         address _to
     ) internal returns (bool) {
-        return IERC20(_neoXToken).transfer(_to, _amount);
+        bytes memory transferCall = abi.encodeCall(IERC20.transfer, (_to, _amount));
+        (bool success, bytes memory returndata) = address(_neoXToken).call(transferCall);
+        return success && (returndata.length == 0 || abi.decode(returndata, (bool))) && address(_neoXToken).code.length > 0;
```
mialbu: Hmm, `address(_neoXToken).code.length > 0` looks interesting here. Is there a specific reasoning for having this check here at the end of the function?

If not, I think it might make more sense if we do such a check outside of this function, i.e., before calling this function in the first place. That way we're also doing the check only once if this function here is called multiple times (assuming `selfdestruct` is not executed in the `_neoXToken` contract during a `transfer` invocation...😅🫣). For example doing this check before the following loop in the `depositToken` function:

https://github.com/bane-labs/bridge-evm-contracts/blob/0c2d7e705a3b8171575942145133a3dfd60d68e7/contracts/bridge/BridgeImpl.sol#L363-L387

and before the execution in a claim:
https://github.com/bane-labs/bridge-evm-contracts/blob/0c2d7e705a3b8171575942145133a3dfd60d68e7/contracts/bridge/BridgeImpl.sol#L421

If you still want to have it directly here in this function, I propose we do it before executing the call and just return `false` if it applies:

```suggestion
        if (address(_neoXToken).code.length == 0) {
                return false;
        }
        bytes memory transferCall = abi.encodeCall(IERC20.transfer, (_to, _amount));
        (bool success, bytes memory returndata) = address(_neoXToken).call(transferCall);
        return success && (returndata.length == 0 || abi.decode(returndata, (bool)));
```

passer-byzhang: You are right, It is better to check outside of this function.
I just rememered that every single `neoXToken` can only be registered by `governor`and every token address is controllable. So I'm not sure it's necessary😵‍💫. If it's neccesary , I will add this check in front of the loop as you say.

mialbu: Afaik, if `_neoXToken` has no deployed code, `success` will be `false` and the `returndata` will be empty after executing the `.call()`. Can you verify that statement?

If so, we **can** add the check as you said outside of this function before it's called. However, I'm not so sure we really need it.

Let's talk about some scenario:
If a registered token address uses `selfdestruct`\*, the worst that can happen, afaik, is that that the relayer keeps invoking `depositToken` for that token address (assuming deposits keep coming in on N3). Ultimately adding all the new deposits to this token bridge's claimable mapping and _paying for the transactions_. However, if there's still deposits coming in on N3, the transactions have been paid for anyway. So I don't really see a considerable risk here. This specific token bridge will go _out of service_ in any way if something like this would happen (and it'll only happen if the token admin decides to destroy the token anyway😅).

So: The scenario seems very unlikely to happen for me AND if it would it won't be a security issue. Hence, I think we don't need the check.

\* We should check if the `_neoXToken` address has code in the registration. I'm actually not sure if we do.

passer-byzhang: Thank you for your analysis!  It makes sense. I will remove this check. Despite that  `success` will be `true` and `returndata` will be empty if the target addresss has no code , this case would never happen because each token should be checked off-chain before `groverner` register it.(such as verifying in explorer or audit report, we can avoid that contract contains a `selfdestruct` in that step).

txhsl (approved):

mialbu (approved):

contracts/bridge/BridgeImpl.sol
```diff
@@ -415,7 +415,8 @@ contract BridgeImpl is
                 tokenType == StorageTypes.TokenType.ERC20Capped
         );
         // Note: For NEO tokens, the transfer value has already been extended with 18 decimals in the deposit function.
-        TokenBridgeLib._executeERC20Transfer(_neoXToken, claimable.amount, to);
+        bool success = TokenBridgeLib._executeERC20Transfer(_neoXToken, claimable.amount, to);
+        if(!success) revert TransferFailed();
```
mialbu: Oh, good catch, that was a bug.

Minor format though:
```suggestion
        bool success = TokenBridgeLib._executeERC20Transfer(_neoXToken, claimable.amount, to);
        if (!success) revert TransferFailed();
```

mialbu (changes requested):

mialbu (approved):

PR #110, opened 2024-05-29
Fix the registered check in function `withdrawToken`.

passer-byzhang:


mialbu (approved):

mialbu:
@passer-byzhang I didn't delete the branch after merging because I don't know what branching model you use. Please delete the branch `passer-byzhang-patch-2` if you no longer need it. (I usually create a new branch for one PR (feature/patch/fix, etc.) and then I delete it again once merged)

PR #118, opened 2024-06-13
Update storage slot allocation

mialbu:
> **🚨 Important 🚨:** Once this PR is merged, a previous commit is required for upgrading the currently deployed bridge-related contracts on T3 in order to circumvent complex migration logic.

The changes in this PR intend to improve the maintainability for potential future upgrades by introducing spacing between the slot assignment of variables.

**Updated storage layout of `BridgeImpl.sol`**
```bash
| Name             | Type                                                                  | Slot | Offset | Bytes | Contract                                   |
|------------------|-----------------------------------------------------------------------|------|--------|-------|--------------------------------------------|
| _status          | uint256                                                               | 0    | 0      | 32    | contracts/bridge/BridgeImpl.sol:BridgeImpl |
| _gap0            | uint256[99]                                                           | 1    | 0      | 3168  | contracts/bridge/BridgeImpl.sol:BridgeImpl |
| management       | contract IBridgeManagement                                            | 100  | 0      | 20    | contracts/bridge/BridgeImpl.sol:BridgeImpl |
| bridgePaused     | bool                                                                  | 100  | 20     | 1     | contracts/bridge/BridgeImpl.sol:BridgeImpl |
| unclaimedRewards | uint256                                                               | 101  | 0      | 32    | contracts/bridge/BridgeImpl.sol:BridgeImpl |
| _gap1            | uint256[97]                                                           | 102  | 0      | 3104  | contracts/bridge/BridgeImpl.sol:BridgeImpl |
| claimableGas     | mapping(uint256 => struct StorageTypes.Claimable)                     | 199  | 0      | 32    | contracts/bridge/BridgeImpl.sol:BridgeImpl |
| tokenBridges     | mapping(address => struct StorageTypes.TokenBridge)                   | 200  | 0      | 32    | contracts/bridge/BridgeImpl.sol:BridgeImpl |
| tokenClaimables  | mapping(address => mapping(uint256 => struct StorageTypes.Claimable)) | 201  | 0      | 32    | contracts/bridge/BridgeImpl.sol:BridgeImpl |
| gasBridge        | struct StorageTypes.GasBridge                                         | 202  | 0      | 320   | contracts/bridge/BridgeImpl.sol:BridgeImpl |
```

**Updated storage layout of `BridgeManagementImpl.sol`**
```bash
| Name               | Type         | Slot | Offset | Bytes | Contract                                                           |
|--------------------|--------------|------|--------|-------|--------------------------------------------------------------------|
| _gap0              | uint256[100] | 0    | 0      | 3200  | contracts/management/BridgeManagementImpl.sol:BridgeManagementImpl |
| owner              | address      | 100  | 0      | 20    | contracts/management/BridgeManagementImpl.sol:BridgeManagementImpl |
| relayer            | address      | 101  | 0      | 20    | contracts/management/BridgeManagementImpl.sol:BridgeManagementImpl |
| validatorThreshold | uint256      | 102  | 0      | 32    | contracts/management/BridgeManagementImpl.sol:BridgeManagementImpl |
| validators         | address[]    | 103  | 0      | 32    | contracts/management/BridgeManagementImpl.sol:BridgeManagementImpl |
| governor           | address      | 104  | 0      | 20    | contracts/management/BridgeManagementImpl.sol:BridgeManagementImpl |
| securityGuard      | address      | 105  | 0      | 20    | contracts/management/BridgeManagementImpl.sol:BridgeManagementImpl |
| funder             | address      | 106  | 0      | 20    | contracts/management/BridgeManagementImpl.sol:BridgeManagementImpl |
```

### DIFF
--- a/contracts/bridge/BridgeStorageV1.sol
+++ b/contracts/bridge/BridgeStorageV1.sol
@@ -3,20 +3,82 @@ pragma solidity ^0.8.24;
 
 import "../interfaces/IBridgeManagement.sol";
 import "../library/StorageTypes.sol";
+import "@openzeppelin/contracts/utils/ReentrancyGuard.sol";
 
-contract BridgeStorageV1 {
+/**
+ * @title SlotShiftV1
+ * @author BaneLabs
+ * @dev This contract shifts the storage slots by 100 of any extending contract - making room for future upgrades.
+ */
+contract SlotShiftV1 {
+    // Slots 0-99 remain empty for future upgrades
+    uint256[100] private _gap0_99;
+}
+
+/**
+ * @title BridgeStorageV1
+ * @author BaneLabs
+ * @dev This contract holds the storage variables for the Bridge contract and outlines the slot allocations of the contracts storage. Storage slot modifications should be done with care to avoid conflicts with existing storage slots in the proxy.
+ */
+contract BridgeStorageV1 is SlotShiftV1, ReentrancyGuard {
+    // Slots 0-99 are taken by SlotShiftV1 (SlotShiftV1 has 100 empty slots)
+    // Slot 100 is taken by ReentrancyGuard's _status storage value
+    // Slots 101-200 remain empty for future upgrades
+    uint256[100] private _gap101_200;
+
+    // ###############################
+    // Contract-wide storage variables
+    // ###############################
+
+    // Slot 201
     IBridgeManagement public management;
-
+    uint96 private _gap201_21_32; // Shifting the bool value 'bridgePaused' to the next slot.
+    // Slot 202
     bool public bridgePaused;
+    // Slot 203
+    uint256 public unclaimedRewards;
+    // Slots 204-300 remain empty for future upgrades
+    uint256[97] private _gap204_300;
 
-    StorageTypes.GasBridge public gasBridge;
+    // ####################################
+    // GasBridge-specific storage variables
+    // ####################################
+
+    // Slot 301
     mapping(uint256 nonce => StorageTypes.Claimable claimable)
         public claimableGas;
+    // Slots 302-311 (10 slots)
+    StorageTypes.GasBridge public gasBridge;
+    // Slots 312-400 remain empty for future upgrades
+    uint256[89] private _gap312_400;
 
-    uint256 public unclaimedRewards;
+    // ######################################
+    // TokenBridge-specific storage variables
+    // ######################################
 
+    // Slot 401
     mapping(address tokenAddress => StorageTypes.TokenBridge tokenBridge)
         public tokenBridges;
+    // Slot 402
     mapping(address tokenAddress => mapping(uint256 nonce => StorageTypes.Claimable claimable) claimableTokens)
         public tokenClaimables;
+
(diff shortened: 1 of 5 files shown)


contracts/bridge/BridgeStorageV1.sol
```diff
@@ -3,20 +3,81 @@ pragma solidity ^0.8.24;
...
+    // Slots 0-99 remain empty for future upgrades
+    uint256[100] private _gap0_99;
+}
+
+/**
+ * @title BridgeStorageV1
+ * @author BaneLabs
+ * @dev This contract holds the storage variables for the Bridge contract and outlines the slot allocations of the contracts storage. Storage slot modifications should be done with care to avoid conflicts with existing storage slots in the proxy.
+ */
+contract BridgeStorageV1 is SlotShiftV1, ReentrancyGuard {
```
chenquanyu: I think we only need gap for abstract contracts and `ReentrancyGuard` can be put in the first slot.

mialbu: I tend to disagree with not needing it. We don't know what a future logic contract will look like, new features, need for more storage slots for one part of the contract logic (it's simpler if vars that belong together are put in following slots for dev purposes instead of entirely different slots). For example, if there's an extension of the GasBridge struct, it'll be a lot simpler, if we can just append it, reduce the followed gap and initialize the new values instead of migrating the complete struct to different slots.

It doesn't infer bigger cost afaik and tbh I'd prefer a bit of storage allocation handling over migrating data in an upgrade any day. It might look a bit over-engineered right now, but I think it's worth it.

mialbu: Do you know of a downside with this change that I don't know of? Except that we need to check what slots the vars are in and keep track of the gap size on upgrades (former we need to do anyway, and latter seems like a rather minor thing imho).

mialbu: If you think it's better, I can remove the first gap (i.e., the `_gap0` in the `BridgeImpl`). For the other gaps, I think it's better to keep them.

chenquanyu: The biggest disadvantage is that it introduces a lot of unnecessary gaps, such as the `_gap0` at the beginning. For upgradeable contracts, the storage area can be added at the end, so the gap at the end is also unnecessary. For struct variables, it is special because it will occupy multiple slots. We can put it at the end, or add new variables when upgrading instead of extending its properties.

mialbu: Okay, I've reduced the use of gaps now. Tbh, I'm very reluctant to reduce it more after commit ebe0a83.

mialbu:
I'll move forward with this. If anyone is really not okay with the allocation, please open a new issue, but for now, I don't want this to block further changes.

contracts/bridge/BridgeStorageV1.sol
mialbu: @chenquanyu I've changed my mind about the gaps **inbetween** variables. But I'm keeping the `_gap0` due to reasons of potentially introducing additional contract extensions that hold variables. That way such vars could take up slots of the gap and other storage vars can be kept in the same file.

I've removed `_gap1` in commit f4a3a39d62cf16e403bf669172d55f3d13e4083c.

PR #119, opened 2024-06-14
Parameter Change Mechanism

### LINKED ISSUE #117: Add Parameter Change Mechanism

mialbu:
The `fee`, `minAmount`, `maxAmount` and `maxDeposits` are considered bridge parameters. These parameters are part of the gas bridge as well as any token bridge. Instead of letting the governor directly change such a parameter in one transaction, I've introduced the concept of a **pending period** and an **execution window**.

If a parameter should be changed, it can be **initiated**. Once initiated, its status is pending, i.e., it is registered and bridge users can see that this change is intended to happen soon. The pending status depends on the block number when it was initiated and the contract-wide `pendingPeriod` integer value. Once the pending period has passed (i.e., the current block number is greater than the pendingUntilBlock value of the initiated change), the change becomes executable.

An executable change is in the **execution window**. The length of this window is defined by the contract-wide `executionWindow` integer value. During the execution window the change can be executed, i.e., its initiated value can be set according to its parameter type.

Once the execution window has passed (i.e., the current block number is greater than the `executableUntilBlock` value of the initiated change), the change is expired and can no longer be executed.

There can only be one active pending change per bridge (i.e., 1 for gas and 1 for each token). An active change can be overwritten only if it's no longer pending. A change is deleted once executed.


txhsl:
Seems like a time lock thing, maybe we can refer to https://docs.openzeppelin.com/contracts/5.x/api/governance.

Let me learn about the concept first. :)

txhsl:
Why do we need this feature? I mean, the bridge setting updates are not something like governance proposals. We only have 1 governor in the contract management, so his decisions have no need to stay pending.

mialbu:
> Why do we need this feature? I mean, the bridge setting updates are not something like governance proposals. We only have 1 governor in the contract management, so his decisions have no need to stay pending.

That is true, the pending part is based on https://github.com/bane-labs/bridge-neo-contracts/issues/47. I'm open to discuss it. If @steven1227 and @gsmachado think we don't need it, we could also ignore this issue.

txhsl:
> That is true, the pending part is based on https://github.com/bane-labs/bridge-neo-contracts/issues/47.

I've read that issue, but as I've commented there, I'm afraid that this PR can not solve the problem as we expect.

mialbu:
We will not move forward with this. #121 will fix the original issue.

PR #121, opened 2024-06-19
Introduce Max Fee and other minor changes

### LINKED ISSUE #120: Add maxFee parameter in withdraw functions

mialbu:
The withdraw functions now expect an additional parameter `_maxFee`. It specifies the maximum fee that the sender is willing to pay for the withdrawal. If the actual fee is higher than this value, the withdrawal is aborted.

Additionally, this PR adds/changes the following:
- More concise errors `AmountBelowMinAmount` and `AmountExceedsMaxAmount`.
- In gas withdrawal, check min and max amount in the total msg.value and not only after the fee is deducted.

### DIFF
--- a/contracts/bridge/BridgeImpl.sol
+++ b/contracts/bridge/BridgeImpl.sol
@@ -443,18 +443,16 @@ contract BridgeImpl is BridgeStorage, IBridge, IGasBridge, ITokenBridge {
     }
 
     /**
-     * @notice Withdraw tokens to Neo N3. Requires that the sender has approved the provided amount to the bridge contract.
+     * @notice Withdraw tokens to Neo N3. Requires that the sender has approved the provided amount to the bridge contract. The fee required for withdrawing that token can be fetched from its config (i.e., tokenBridges[_neoXToken].config.fee). It needs to be payed to this function (i.e., as msg.value).
      * @dev This function transfers the provided amount of the provided token from the msg.sender to this contract. It requires that the msg.sender has previously approved at least the provided amount to this contract. Further, it computes the new root and updates the token withdrawal state.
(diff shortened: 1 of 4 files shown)


contracts/bridge/BridgeImpl.sol
```diff
@@ -455,9 +466,14 @@ contract BridgeImpl is BridgeStorage, IBridge, IGasBridge, ITokenBridge {
...
-        uint256 tokenValue = _amount;
-        if (tokenValue < config.minAmount) revert InvalidAmount();
-        if (tokenValue > config.maxAmount) revert InvalidAmount();
+        uint256 tokenAmount = _amount;
+        if (tokenAmount < config.minAmount)
+            revert AmountBelowMinAmount(config.minAmount, tokenAmount);
+        if (tokenAmount > config.maxAmount)
+            revert AmountExceedsMaxAmount(config.maxAmount, tokenAmount);
+        // Revert if the actual fee is higher than the provided max fee.
+        if (config.fee > _maxFee) revert MaxFeeExceeded(_maxFee, config.fee);
```
txhsl: Well, that's a bit reduplicated. The `msg.value` works the same as `_maxFee`. If I declare a higher `_maxFee` than I pay, then the declaration won't work. If I declare a lower `_maxFee` than I pay, then the left will never return.

How about remove `_maxFee` in `withdrawToken`?

mialbu: You're absolutely right. The `maxFee` logic is redundant in `withdrawToken`. I'll remove it.

mialbu: Removed in 1b7b6ed146018632de8ec942dd65e3fcda70e09c.

txhsl (approved):

PR #133, opened 2024-07-03
Use actual balance increase as amount in token withdrawal

### LINKED ISSUE #128: [S3A Medium] Fix inconsistent balance on transfer-on-fee or deflationary tokens
### Context
https://github.com/Secure3Audit/findings_NeoX_Bridge_Contract/issues/22

mialbu:
Resolves #128 

### Context

Based on a raised issue in the security audit (https://github.com/Secure3Audit/findings_NeoX_Bridge_Contract/issues/22), there might be tokens that are designed in a way that they deduct a fee of the to-be-transferred token, leading to a smaller amount received than sent. This PR checks the bridge contract's balance before and after the transfer execution and takes into account the actually received amount for the bridging operation.

Here's part of the statement in the issue:

> This function is used to withdraw tokens to Neo N3 and it requires that the sender has approved the provided amount to the bridge contract, issue however is that when transferring the execution uses the classic IERC20.transferFrom(), however for popular tokens that deduct fees on their transfers, this then causes protocol to not only inflate the amount of tokens received, but some invariants could also be broken.
>
> POC for broken invariant:
>
> Here we are to prove a user can deposit < config.minAmount.
> config.minAmount for the token is set as 10e18.
> User calls on withdraw with 106e17.
> The token's fee for transfers is 10%
> User ends up depositing ~ 95e17 which is < minAmount = 10e18
> As shown by the minimalistic POC, not only is the accounting flawed, but users could as well even break an invariant.

### DIFF
--- a/contracts/bridge/BridgeImpl.sol
+++ b/contracts/bridge/BridgeImpl.sol
@@ -485,6 +485,8 @@ contract BridgeImpl is BridgeStorage, IBridge, IGasBridge, ITokenBridge {
 
         // Compare the balance before and after the transfer to get the actual received amount. This is necessary if the token contract were to deduct a fee in transfers.
         uint256 bridgeBalanceAfter = erc20Token.balanceOf(address(this));
+        // Revert if there is an underflow.
+        if (bridgeBalanceAfter < bridgeBalanceBefore) revert InvalidTransfer();
         uint256 receivedAmount = bridgeBalanceAfter - bridgeBalanceBefore;
 
         // Check that the received amount is in the allowed range.
--- a/contracts/bridge/BridgeImpl.sol
+++ b/contracts/bridge/BridgeImpl.sol
@@ -447,10 +447,10 @@ contract BridgeImpl is BridgeStorage, IBridge, IGasBridge, ITokenBridge {
 
     /**
      * @notice Withdraw tokens to Neo N3. Requires that the sender has approved the provided amount to the bridge contract. The fee required for withdrawing that token can be fetched from its config (i.e., tokenBridges[_neoXToken].config.fee). It needs to be payed to this function (i.e., as msg.value).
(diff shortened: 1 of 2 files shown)


contracts/bridge/BridgeImpl.sol
```diff
@@ -485,23 +483,33 @@ contract BridgeImpl is BridgeStorage, IBridge, IGasBridge, ITokenBridge {
         );
         if (!success) revert TransferFailed();
 
+        // Compare the balance before and after the transfer to get the actual received amount. This is necessary if the token contract were to deduct a fee in transfers.
+        uint256 bridgeBalanceAfter = erc20Token.balanceOf(address(this));
+        uint256 receivedAmount = bridgeBalanceAfter - bridgeBalanceBefore;
```
txhsl, line 490: Just an assumption, since we suppose that anything can happen in `IERC20(_neoXToken).transferFrom`, then if `bridgeBalanceAfter` is smaller than `bridgeBalanceBefore`, there will be an underflow.

mialbu: Yes, you're right. I'm thinking of adding a check for this before the subtraction. For example:

`if (bridgeBalanceAfter < bridgeBalanceBefore) revert InvalidTransfer();`

What do you think about this? Do you have a different solution in mind?

txhsl: > I'm thinking of adding a check for this before the subtraction.

It's OK, just compare before sub. https://github.com/OpenZeppelin/openzeppelin-contracts/blob/release-v5.0/contracts/utils/math/Math.sol#L36

mialbu: Fixed in [560679a](https://github.com/bane-labs/bridge-evm-contracts/pull/133/commits/560679a0afc80de70eafc59085222675b307b232).

txhsl (approved):

PR #135, opened 2024-07-03
Introduce 2-step ownership transfer

### LINKED ISSUE #134: [S3A Low] Use two-step process to transfer ownership
### Context
https://github.com/Secure3Audit/findings_NeoX_Bridge_Contract/issues/33

mialbu:
Resolves #134.

### DIFF
--- a/contracts/management/BridgeManagementImpl.sol
+++ b/contracts/management/BridgeManagementImpl.sol
@@ -62,13 +62,8 @@ contract BridgeManagementImpl is BridgeManagementStorage, IBridgeManagement {
         return covered == threshold;
     }
 
-    function setOwner(address _owner) external onlyOwner {
-        _setOwner(_owner);
-        emit OwnerChange(_owner);
-    }
-
     function getOwner() external view override returns (address) {
-        return owner;
+        return owner();
     }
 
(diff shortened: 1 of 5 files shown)


contracts/management/BridgeManagementImpl.sol
```diff
@@ -62,13 +62,8 @@ contract BridgeManagementImpl is BridgeManagementStorage, IBridgeManagement {
         return covered == threshold;
     }
 
-    function setOwner(address _owner) external onlyOwner {
-        _setOwner(_owner);
-        emit OwnerChange(_owner);
-    }
-
     function getOwner() external view override returns (address) {
```
txhsl: Maybe we don't need this `getOwner()` as well, I think `owner()` is publicly accessible.

mialbu: I agree. However, when I wanted to remove it I needed to change the interface, then the Bridge contract and then I wanted to change other getters as well to stay consistent, i.e., remove other get prefixes as well ('getRelayer()' to 'relayer()').

I'll change it for the owner here and consider removing the "get" prefixes for other role getters in another PR.

txhsl (approved):

PR #143, opened 2024-07-09
Fix requirements for min and max amount

### LINKED ISSUE #142: [S3A Low] `maxAmount` should be strictly greater than `minAmount`
### Context

https://github.com/Secure3Audit/findings_NeoX_Bridge_Contract/issues/64

mialbu:
Fixes #142

Additionally to #142, a redundant check was removed, since the case is already covered in the underlying call to `_isValidConfig()`.

### DIFF
--- a/contracts/bridge/BridgeStorage.sol
+++ b/contracts/bridge/BridgeStorage.sol
@@ -253,7 +253,7 @@ abstract contract BridgeStorage is BridgeStorageV1, UUPSUpgradeable {
         address _neoXToken,
         uint256 _amount
     ) internal {
-        if (_amount > tokenBridges[_neoXToken].config.maxAmount)
+        if (_amount >= tokenBridges[_neoXToken].config.maxAmount)
             revert InvalidAmount();
         tokenBridges[_neoXToken].config.minAmount = _amount;
     }
@@ -262,7 +262,7 @@ abstract contract BridgeStorage is BridgeStorageV1, UUPSUpgradeable {
         address _neoXToken,
         uint256 _amount
     ) internal {
-        if (_amount < tokenBridges[_neoXToken].config.minAmount)
+        if (_amount <= tokenBridges[_neoXToken].config.minAmount)
             revert InvalidAmount();
         tokenBridges[_neoXToken].config.maxAmount = _amount;
     }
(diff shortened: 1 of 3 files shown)


contracts/bridge/BridgeStorage.sol
```diff
@@ -253,7 +253,7 @@ abstract contract BridgeStorage is BridgeStorageV1, UUPSUpgradeable {
         address _neoXToken,
         uint256 _amount
     ) internal {
-        if (_amount > tokenBridges[_neoXToken].config.maxAmount)
+        if (_amount >= tokenBridges[_neoXToken].config.maxAmount)
```
txhsl, line 256: This means the `_amount` can not equal `maxAmount`, which seems not user-friendly, and it doesn't help #142. Is it what we expect?

mialbu: How does it not help #142?

Check out the S3A issue [here](https://github.com/Secure3Audit/findings_NeoX_Bridge_Contract/issues/64).

If we have a `min` and a `max` I agree with the issue in the sense that it's generally not good practice to allow them to be equal. Also, the behavior in the `registerToken()` function was already restricting it (as well as the corresponding functions on the N3 bridge contract), while the setters for the min and max amounts had other logic, i.e., allowing the values to be equal. This means that the current logic in the setters is inconsistent with `registerToken()` and the logic on the N3 side.

mialbu: I'm not completely against allowing them to be equal, but **if** we were to allow them to be equal, we need to change the logic consistently in the registration, the setters and on the N3 side.

mialbu: Okay, just checked, the N3 side misses the check entirely in the setters. 😂 I'll open an issue to fix it.

txhsl: > How does it not help https://github.com/bane-labs/bridge-evm-contracts/issues/142?
> Check out the S3A issue https://github.com/Secure3Audit/findings_NeoX_Bridge_Contract/issues/64.

My bad, it helps.

mialbu: Opened an issue for it: https://github.com/bane-labs/bridge-neo-contracts/issues/109.

txhsl (approved):

Issue #130, opened 2024-07-02
[S3A Low] Use `safeTransferFrom` instead of `transferFrom`


mialbu:
### Context
- https://github.com/Secure3Audit/findings_NeoX_Bridge_Contract/issues/57
- https://github.com/Secure3Audit/findings_NeoX_Bridge_Contract/issues/58
- https://github.com/Secure3Audit/findings_NeoX_Bridge_Contract/issues/70

mialbu:
Fixed in #173

PR #171, opened 2024-07-16
Refund excess `msg.value`

### LINKED ISSUE #160: [S3A Informational] Return excess `msg.value`
### Context

https://github.com/Secure3Audit/findings_NeoX_Bridge_Contract/issues/16

### LINKED ISSUE #161: [S3A Informational] Consistently use config.fee for unclaimedRewards
### Context

https://github.com/Secure3Audit/findings_NeoX_Bridge_Contract/issues/17

mialbu:
Fixes #160 
Fixes #161

I ran into a `stack too deep` compilation error. Therefore, I had to extract some code from the `withdrawToken()` function, i.e., the ERC20 transfer is now in a separate function `_transferERC20TokenToBridge()`.

### DIFF
--- a/contracts/bridge/BridgeImpl.sol
+++ b/contracts/bridge/BridgeImpl.sol
@@ -476,31 +476,20 @@ contract BridgeImpl is BridgeStorage, IBridge, IGasBridge, ITokenBridge {
         StorageTypes.TokenConfig memory config = _getTokenConfig(_neoXToken);
 
         // Revert if the provided value is lower than the required fee.
-        if (msg.value < config.fee)
-            revert InsufficientFee(config.fee, msg.value);
-        _addUnclaimedRewards(msg.value);
+        uint256 fee = config.fee;
+        if (msg.value < fee) revert InsufficientFee(fee, msg.value);
+        // Refund the sender if the provided value is higher than the required fee.
+        if (msg.value > fee) {
+            _refund(msg.sender, msg.value - fee);
+        }
+        _addUnclaimedRewards(fee);
 
-        IERC20 erc20Token = IERC20(_neoXToken);
-        uint256 bridgeBalanceBefore = erc20Token.balanceOf(address(this));
-        // Execute the transfer of the tokens from the sender to the bridge contract.
-        bool success = IERC20(_neoXToken).transferFrom(
-            msg.sender,
-            address(this),
-            _amount
+        uint256 receivedAmount = _executeERC20Transfer(
+            _neoXToken,
+            _amount,
+            config.minAmount,
+            config.maxAmount
         );
-        if (!success) revert TransferFailed();
-
-        // Compare the balance before and after the transfer to get the actual received amount. This is necessary if the token contract were to deduct a fee in transfers.
-        uint256 bridgeBalanceAfter = erc20Token.balanceOf(address(this));
-        // Revert if there is an underflow.
-        if (bridgeBalanceAfter < bridgeBalanceBefore) revert InvalidTransfer();
-        uint256 receivedAmount = bridgeBalanceAfter - bridgeBalanceBefore;
-
-        // Check that the received amount is in the allowed range.
-        if (receivedAmount < config.minAmount)
-            revert AmountBelowMinAmount(config.minAmount, receivedAmount);
-        if (receivedAmount > config.maxAmount)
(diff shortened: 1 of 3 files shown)


contracts/bridge/BridgeImpl.sol
```diff
@@ -537,6 +531,29 @@ contract BridgeImpl is BridgeStorage, IBridge, IGasBridge, ITokenBridge {
...
 
+    function _transferERC20TokenToBridge(
+        address _neoXToken,
+        address _from,
+        uint256 _amount
+    ) private returns (uint256 actualReceivedAmount) {
+        IERC20 erc20Token = IERC20(_neoXToken);
+        uint256 bridgeBalanceBefore = erc20Token.balanceOf(address(this));
+        // Execute the transfer of the tokens from the sender to the bridge contract.
+        bool success = IERC20(_neoXToken).transferFrom(
```
txhsl, line 574: Use `SafeERC20.safeTransferFrom()` here as well.

contracts/bridge/BridgeImpl.sol
```diff
@@ -476,26 +476,20 @@ contract BridgeImpl is BridgeStorage, IBridge, IGasBridge, ITokenBridge {
...
 
         // Revert if the provided value is lower than the required fee.
-        if (msg.value < config.fee)
-            revert InsufficientFee(config.fee, msg.value);
-        _addUnclaimedRewards(msg.value);
+        uint256 fee = config.fee;
+        if (msg.value < fee) revert InsufficientFee(fee, msg.value);
+        // Refund the sender if the provided value is higher than the required fee.
+        if (msg.value > fee) {
+            (bool success, ) = _to.call{value: msg.value - fee}("");
```
txhsl: What is the reason that we use `_to.call()` here?

First, the address `_to` is the N3 target address instead of the `msg.sender`.

Second, the auditor mentioned `payable(msg.sender).transfer(msg.value - config.fee);`, and I think it is somehow reasonable. Here I don't expect to trigger any `payable` contract logic, because it looks quite tricky. `transfer()` only gives 2300 gas, otherwise revert.

Well, but obviously, `transfer()` is not the best choice as well. How do you think?

txhsl: At least, let's put this refund at the end of withdraw.

txhsl (changes requested):

contracts/bridge/BridgeImpl.sol
mialbu: This will be addressed in the scope of #130.

contracts/bridge/BridgeImpl.sol
mialbu: > First, the address `_to` is the N3 target address instead of the `msg.sender`.

🫣 Good catch!

mialbu: > Well, but obviously, `transfer()` is not the best choice as well. How do you think?

Yes, `transfer()` is not nice due to potential changes in the future that could change the required gas for such an execution (a scenario where 2300 gas is not sufficient). However, if there needs to be a refund, we can also just check if it's a contract and revert if it is.

So, for example: If a refund is needed, execute it if the msg.sender is an EOA. Otherwise, revert the complete transaction.

txhsl: > So, for example: If a refund is needed, execute it if the msg.sender is an EOA. Otherwise, revert the complete transaction.

Will this be added in this PR?

mialbu: > If a refund is needed, execute it if the msg.sender is an EOA. Otherwise, revert the complete transaction.

Yes, this part would be in this PR. What do you think about it?

txhsl: It's good if possible.

mialbu: It is possible. ☺️

mialbu: We do the same here, when the relayer executes a gas deposit:

https://github.com/bane-labs/bridge-evm-contracts/blob/0e3ad1c40c10877e5446ba350a09207cc08bea5b/contracts/bridge/BridgeImpl.sol#L136-L139

Gas deposits to contracts are added to the claimable gas mapping, in order for the relayer not to pay and trigger uncertain payable contract logic.

txhsl (approved):

PR #175, opened 2024-07-18
Release `0.5.0`

mialbu:
@txhsl and @chenquanyu

Please take a careful look at the release files. For compiling the release artifacts, I have done some adaptions according to the requirements of the corresponding network and its state (precompile for mainnet vs. upgrade in T4). Please read the following changes and provide feedback if something should be unclear, not right or missing.

### Mainnet

_m-1_
I removed the `initialize` function for compiling the byte code for the mainnet genesis file. The `initializer` function will never and can never actually be executed, since the initial state is already set in the genesis file (as is the initialization slot, which is set to 1 in the genesis file)

### Testnet

_t-1_
I removed the `initialize` function since the v1 logic contract on T4 is already properly initialized and the `initialize` function cannot be called again anyway. A further reinitialization is not necessary as this version upgrade does not require any storage change. Hence, a function with a `reinitializer(2)` modifier is not necessary.

_t-2_
I removed the constant `SELF` and the overridden functions `_checkProxy()` and `_checkNotDelegated()` since this new logic contract will actually go through the deployment process and the logic contract's `immutable __self` variable will be set correctly and the default `_checkProxy()` and `_checkNotDelegated()` functions will take the expected effect.


txhsl (approved):
LGTM.

PR #181, opened 2024-08-12
Remove functions specifically required for precompiled version

mialbu:
The removed functions were specifically required for the precompile versions since they override functions that consider an immutable (set during deployment) variable of the logic contract. Because the deployment part is not executed for the precompile version, these needed to be overridden with specific data suited for the precompile.

Thus, now for a logic update, a new logic contract needs to be deployed. This means its deployment part will be executed and its immutable variable will be set, so that the original functions must be used.

### DIFF
--- a/contracts/bridge/BridgeStorage.sol
+++ b/contracts/bridge/BridgeStorage.sol
@@ -365,37 +365,4 @@ abstract contract BridgeStorage is BridgeStorageV1, UUPSUpgradeable {
     function _authorizeUpgrade(
         address newImplementation
     ) internal virtual override onlyAdmin {}
-
-    // UUPSUpgradeable-specific functions required for precompiled version.
-
-    /**
-     * @dev Reverts if the execution is not performed via delegatecall or the execution
-     * context is not of a proxy with an ERC-1967 compliant implementation pointing to self.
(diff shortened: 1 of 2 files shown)


txhsl:
The following two can be removed as well, for the same reason.
https://github.com/bane-labs/bridge-evm-contracts/blob/2432166d7d546380aaad6aa8c3973ef27c32c3c2/contracts/bridge/BridgeStorage.sol#L16
https://github.com/bane-labs/bridge-evm-contracts/blob/2432166d7d546380aaad6aa8c3973ef27c32c3c2/contracts/management/BridgeManagementStorage.sol#L14

mialbu:
> The following two can be removed as well, for the same reason.
> 
> https://github.com/bane-labs/bridge-evm-contracts/blob/2432166d7d546380aaad6aa8c3973ef27c32c3c2/contracts/bridge/BridgeStorage.sol#L16
> 
> 
> https://github.com/bane-labs/bridge-evm-contracts/blob/2432166d7d546380aaad6aa8c3973ef27c32c3c2/contracts/management/BridgeManagementStorage.sol#L14

Removed in commit https://github.com/bane-labs/bridge-evm-contracts/pull/181/commits/920f797c6e7639179f07a3dd35bc91dc4e1ca782.

txhsl (approved):

PR #189, opened 2024-10-04
Add scripts for simple deployment of bridge for testing purposes

mialbu:
FYI: I'm using the base branch `1.0.0-compatible/dev` (branched from `main`) to add code in the repo that is compatible with the contract versions 1.0.0, i.e., the contract states of the contracts that are currently deployed.


mialbu:
This PR is for v1.0.0 compatibility. If we're merging #192 and upgrade the contracts, this PR becomes redundant and I'll close it.

PR #193, opened 2024-10-08
Add decimal scaling factor

mialbu:
Previously the variable `ExecutionType` was used for deciding if the amount used in the hashing computation should be used for the transfer amount or if 18 decimals should be added to it (i.e., for NEO, since it has no decimals on N3, but likely will have 18 decimals on Neo X). This restricted the registration of tokens that have different decimal places on N3 and Neo X (other than the difference being exactly 18) and would have required a new execution type for each decimal difference.

This PR replaces the concept of "execution type" with a decimal scaling factor. That decimal scaling factor defines what number of decimals need to be **added** to the amount used for the hash chain to get the actual transfer amount.

The value needs only be set in token registrations if tokens have different decimals on both chains and it only needs to be set on the chain where the token has more decimals since the amount with less decimals is used for the hash chain computation.

For example, with Neo, the decimal scaling factor will need to be 18 on Neo X (given the Neo token on Neo X has 18 decimals (18 more than on N3)) and the decimal scaling factor on the registered Neo token bridge on N3 should be 0.

### DIFF
--- a/contracts/library/StorageTypes.sol
+++ b/contracts/library/StorageTypes.sol
@@ -36,17 +36,6 @@ library StorageTypes {
 
     // Token Bridges
 
-    /**
-     * The execution type is used to specifies the behaviour when executing a token distribution from another chain.
-     * Currently, there's only two types of a token distribution, which both use a simple ERC20's "transfer(address,uint256)" function, while the type NEO additionally adds 18 decimal places when depositing, since Neo does not have decimal places on Neo N3.
-     * How the distribution handling for other types will be implemented is still under discussion. However, for ERC721, the functions in mind are "mint(address,uint256)", "safeMint(address,uint256)", "transferFrom(address,uint256)" or "safeTransferFrom(address,uint256)". As seen from these function declarations, similar to ERC20, they all use the same parameter types, i.e., besides the function call, the complete existing execution logic and chain computation can be reused in depositToken().
-     */
-    enum ExecutionType {
-        NEO,
-        ERC20
-        // ERC721
-    }
-
     struct TokenBridge {
         bool paused;
         State depositState;
@@ -56,11 +45,12 @@ library StorageTypes {
 
     struct TokenConfig {
         address neoN3Token;
+        // The decimal scaling factor should be used if the token on this chain has more decimal precision than the token on the other chain.
+        // For example, if the token on this chain has 18 decimals and the token on the other chain has 8 decimals, the decimal scaling factor should be 10.
+        uint8 decimalScalingFactor;
         uint256 fee;
         uint256 minAmount;
         uint256 maxAmount;
         uint256 maxDeposits;
-        // Execution details
-        ExecutionType executionType;
     }
 }
--- a/contracts/library/StorageTypes.sol
+++ b/contracts/library/StorageTypes.sol
@@ -45,12 +45,12 @@ library StorageTypes {
 
     struct TokenConfig {
         address neoN3Token;
-        // The decimal scaling factor should be used if the token on this chain has more decimal precision than the token on the other chain.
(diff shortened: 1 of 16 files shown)


contracts/library/StorageTypes.sol
```diff
@@ -56,11 +45,12 @@ library StorageTypes {
 
     struct TokenConfig {
         address neoN3Token;
+        // The decimal scaling factor should be used if the token on this chain has more decimal precision than the token on the other chain.
+        // For example, if the token on this chain has 18 decimals and the token on the other chain has 8 decimals, the decimal scaling factor should be 10.
+        uint8 decimalScalingFactor;
```
txhsl: Although I think this is a safe change in storage, how should we handle existing token configurations with `executionType`?

txhsl: OK, so the old `ExecutionType` storage and the new `uint8` are actually at the same place, right?

mialbu: This new value will be in the same storage slot as the appropriate config's `neoN3Token` address, but with an offset of 20. So, for existing tokens this value will just be zero (1). As for the execution type that I removed, that value won't be accessed (2) because it's the last entry of a token bridge struct stored in storage.

- For (1), we should properly initialize this value for existing token bridges.
- (2) needs to be handled as well by e.g., setting their value back to 0. Otherwise, if we add new values to the TokenConfig in the future, this storage slot might already have a non-zero value.

---

> OK, so the old `ExecutionType` storage and the new `uint8` are actually at the same place, right?

No, before it was:
```
slot 1: neoN3Token
slot 2-5: fee, min/maxAmount, maxDeposits
slot 6: executionType
```

Now it's:
```
slot 1: neoN3Token
slot 1 (offset 20\*): decimalScalingFactor
slot 2-5: fee, min/maxAmount, maxDeposits
slot 6: -
```
\* because the address length is 20, the `neoN3Token` only needs the first 20 bytes out of the 32 bytes slot size.

mialbu: We could move the `decimalScalingFactor` to the same slot and use a uint256 for it. However, we'll need a reinitialization with migration logic as well.

I used a `uint8` because we'll never need more than a `uint8` and the [ERC20](https://github.com/ethereum/ercs/blob/master/ERCS/erc-20.md) uses a `uint8` as return value for decimals.

txhsl: > We could move the `decimalScalingFactor` to the same slot and use a uint256 for it. However, we'll need a reinitialization with migration logic as well.

> For (1), we should properly initialize this value for existing token bridges.

Yeh, but I think both of them require a reinitialization. If so, then I prefer to reuse `ExecutionType` storage instead of leaving them there with some non-zero values.

mialbu: > Yeh, but I think both of them require a reinitialization. If so, then I prefer to reuse `ExecutionType` storage instead of leaving them there with some non-zero values.

With "both need reinitialization" I didn't mean that we should just leave the execution type. We definitely should not just leave it there. We should delete the value, setting it to 0.

What we can do is:
1) Use the slot assignment of this PR state now: 
- Reinitialization: set the decimal scaling factor and wipe (set to zero) the execution type in the reinitialization
2) Use the former execution type slot for the decimal scaling factor
- Reinitialization: Set the decimal scaling factor overwriting the former value in the slot

If we overwrite the former execution type slot (2), then we can also just use a uint256 for the decimal scaling factor.

Maybe it's more concise if we just reuse the slot. I'm not opinionated.

txhsl: I'm OK with both.

txhsl (approved):

mialbu:
@txhsl, fyi I'm reusing the same slot now. I think it's the simpler approach.

PR #195, opened 2024-10-10
Move initializer functions to test contracts

mialbu:
The initializer functions are not needed for actual deployment, since the contracts in production are already initialized. I moved the initializer functions to corresponding test contract instances that extend the bridge and management contracts.

If we need to reinitialize, we need a different method with a `reinitializer(newVersion)` modifier and not the `initializer` modifier.

### DIFF
--- a/contracts/tests/TestBridge.sol
+++ b/contracts/tests/TestBridge.sol
@@ -8,13 +8,7 @@ contract TestBridge is BridgeImpl {
     /// @custom:oz-upgrades-unsafe-allow constructor
     constructor() BridgeImpl() {}
 
-    function initialize(
-        address _management,
-        uint256 _fee,
-        uint256 _minAmount,
-        uint256 _maxAmount,
-        uint256 _maxDeposits
-    ) public initializer {
+    function initialize(address _management) public initializer {
         __ReentrancyGuard_init();
         management = IBridgeManagement(_management);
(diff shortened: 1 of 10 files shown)


contracts/tests/TestBridge.sol
```diff
@@ -5,6 +5,31 @@ import "../bridge/BridgeImpl.sol";
...
 contract TestBridge is BridgeImpl {
+    /// @custom:oz-upgrades-unsafe-allow constructor
+    constructor() BridgeImpl() {}
+
+    function initialize(
+        address _management,
+        uint256 _fee,
+        uint256 _minAmount,
+        uint256 _maxAmount,
+        uint256 _maxDeposits
```
txhsl: To make things much simpler, I'm wondering if we can replace some of these input parameters with fixed numbers.

If they're not going to be changed between our tests, then we don't need to leave them changeable as inputs.

mialbu: Yes, we can. The values used in our tests are always the same initially except for the management address ofc.

mialbu: Removed in [2545f7a](https://github.com/bane-labs/bridge-evm-contracts/pull/195/commits/2545f7adfcfa11a9fdaaedbdf2cd5004979f76a3).

txhsl (approved):

PR #196, opened 2024-10-10
Prepare structure for reinitializer logic

mialbu:
The added contracts `BridgeImplV1ToV2` and `BridgeManagementImplV1ToV2` are intended to be the ones that should be deployed once we're upgrading the bridge contracts. They extend the corresponding `...Impl` contracts and hold the migration logic.

I split the migration logic from the `...Impl` contracts, so that we can maintain reproducible steps in the tests later.\*

\* In a similar pattern as when I moved the initializer functions to test contracts to mock the native contract "deployment", I intend to add this reinitializer parts to V2 test contracts that we can then just upgrade the proxy in the tests triggering the reinitializer instead of writing a new initializer just for the tests.

### DIFF
--- a/contracts/bridge/BridgeImplV1ToV2.sol
+++ b/contracts/bridge/BridgeImplV1ToV2.sol
@@ -1,0 +1,28 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.8.25;
+
+import "./BridgeImpl.sol";
+
+contract BridgeImplV1ToV2 is BridgeImpl {
+    /// @custom:oz-upgrades-unsafe-allow constructor
+    constructor() BridgeImpl() {}
+
+    struct TokenMigrationV1 {
+        address token;
+        uint256 decimalScalingFactor;
+    }
+
+    function reinitialize(
+        TokenMigrationV1[] calldata _tokenBridgeMigrations
+    ) public reinitializer(2) {
+        for (uint256 i = 0; i < _tokenBridgeMigrations.length; i++) {
+            TokenMigrationV1 memory migration = _tokenBridgeMigrations[i];
+            if (!_isRegisteredToken(migration.token))
+                revert("Token not registered");
+            StorageTypes.TokenConfig storage config = tokenBridges[
+                migration.token
+            ].config;
+            config.decimalScalingFactor = migration.decimalScalingFactor;
+        }
+    }
+}
--- a/contracts/bridge/BridgeImplV1ToV2.sol
+++ b/contracts/bridge/BridgeImplV1ToV2.sol
@@ -12,9 +12,15 @@ contract BridgeImplV1ToV2 is BridgeImpl {
         uint256 decimalScalingFactor;
     }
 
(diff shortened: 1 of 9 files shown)


txhsl (commented):
`reinitializer` in OpenZeppelin `Initializable.sol` doesn't have any kind of access control, which means, the first guy who calls and provides `_tokenBridgeMigrations` determines the result of reinitialization.

I know it is quite rare to take place, only if somebody is watching our bridge. And it only needs us another upgrade to reinitialize again, without safety concern, if the bridge gets properly paused. But it would be nice if we get this improved.

contracts/bridge/BridgeImplV1ToV2.sol
```diff
@@ -0,0 +1,28 @@
...
+
+    struct TokenMigrationV1 {
+        address token;
+        uint256 decimalScalingFactor;
+    }
+
+    function reinitialize(
+        TokenMigrationV1[] calldata _tokenBridgeMigrations
+    ) public reinitializer(2) {
+        for (uint256 i = 0; i < _tokenBridgeMigrations.length; i++) {
```
txhsl, line 24: The N3 version requires full migration on all tokens, but here is not, ref https://github.com/bane-labs/bridge-neo-contracts/pull/136/files#diff-34e759c10848e46962adf766db7ae6b390eb111a19e9eefd9c345c307331c3bfR17.

mialbu: This is a very valid point and I had this in my mind already. In the current design, the token bridges are only held in a mapping. There's no way of knowing which tokens are registered without external inputs.

I created an [issue](https://github.com/bane-labs/bridge-evm-contracts/issues/198) now that should fix this.

txhsl (approved):

txhsl (commented):
Can we rework 9388bc40ebd62fee6a49362165a0cdfbf8334469 based on https://github.com/bane-labs/bridge-evm-contracts/pull/197#discussion_r1797618898? The point is to have only V2 contract.

I don't think we should have version-based inheritance, we are trying to add new things to `V1`, but deploy `V1ToV2` which is actually `V2`. The definition `contract TestManagementV2 is BridgeManagementImplV1ToV2` makes me confirmed that we need a better structure management.

The `V1ToV2` naming and inheritance are quite misleading to me, e.g. the `contract BridgeImplV1ToV2 is BridgeImpl` declaration, updating `contract BridgeImpl` to `contract BridgeImplV2` is good enough.

mialbu:
I agree with this. Check my [comment](https://github.com/bane-labs/bridge-evm-contracts/pull/197#discussion_r1799045222).

I think we should go forward with this PR, if it's just the structure and fix the structure in #197 or even in another separate PR.

txhsl:
> I think we should go forward with this PR, if it's just the structure and fix the structure in #197 or even in another separate PR.

The structure and tests, I'm not insisting fixing it here, we can go forward.

PR #197, opened 2024-10-10
Fix validator logic

### LINKED ISSUE #165: Support add, remove and replace functionality for validators
### Context

This issue was created in relation to the following acknowledged audit report issue: https://github.com/Secure3Audit/findings_NeoX_Bridge_Contract/issues/79

### LINKED ISSUE #166: Use mapping for validators
### Context

This issue is related to the following acknowledged audit report issue: https://github.com/Secure3Audit/findings_NeoX_Bridge_Contract/issues/2

mialbu:
Closes #166
Resolves #165


contracts/management/BridgeManagementStorageV1.sol
```diff
@@ -26,4 +26,7 @@ abstract contract BridgeManagementStorageV1 is Ownable2StepUpgradeable {
 
     // Slot 105
     address internal funder;
+
+    // Slot 106
+    mapping(address validator => bool isValidator) internal validatorMap;
```
txhsl: I would like to recommend https://github.com/OpenZeppelin/openzeppelin-contracts/blob/release-v5.0/contracts/utils/structs/EnumerableSet.sol, but there seems some problems to inherit the current storage.

But what about a similar `address=>position` mapping?

contracts/management/BridgeManagementStorage.sol
```diff
@@ -15,29 +15,72 @@ abstract contract BridgeManagementStorage is
...
+        if (validators[_index] != _oldValidator)
+            revert IndexValidatorMismatch(validators[_index], _oldValidator);
+        validatorMap[_oldValidator] = false;
+        validatorMap[_newValidator] = true;
+        validators[_index] = _newValidator;
+    }
+
+    function _setValidatorThreshold(uint256 _threshold) internal {
+        if (_threshold <= 1) revert InvalidValidatorThreshold();
+        if (_threshold > validators.length) revert InvalidValidatorThreshold();
```
txhsl: Conditions could merge.

contracts/tests/TestBridgeManagement.sol
```diff
@@ -24,8 +29,26 @@ contract TestBridgeManagement is BridgeManagementImpl {
         _setFunder(_funder);
     }
 
-    // Authorize the contract owner to upgrade the contract for testing purposes.
-    function _authorizeUpgrade(
-        address newImplementation
-    ) internal virtual override onlyOwner {}
+    function _setValidators(
```
txhsl: Do we still need this? For V1-to-V2 migration testing?

If the migration works and the contract gets upgraded finally, then I expect the V1 related things get removed. We need to avoid having V1, V1ToV2, V2 at the same time, but only one `BridgeManagement` in this repo. We can leave some comments about where is the history V1 code release, or why the method `upgradeToV2()` is necessary.

For a migration UT, I think a simplest mock contract with `validators` storage initialized is enough. About cheat codes for storage, please refer https://book.getfoundry.sh/reference/forge-std/std-storage.

mialbu: This sounds simpler than it actually is. I would not work with storage slot allocations outside of solidity code. Nevertheless, I agree with you, the naming and the current structure is confusing.

I think, there should be no V1ToV2. The reinitializer should be in V2 directly. When using it in tests, we can just not call the reinitializer function.

mialbu: Also, we currently have tests using foundry and hardhat and we'd need to maintain both of them with correct storage allocation. I'd rather keep some functionality in a Test-contract that extends the development version of the contract and contains some logic to set storage slots to "mock" the current slot allocation.

We need such a Test-contract anyway to override the `onlyOwner` modifier for example.

mialbu: But I'll get rid of V1ToV2 and just put the reinitializer logic directly in the Impl contract.

mialbu: @txhsl I removed V1ToV2. Check the commits [b4a2487](https://github.com/bane-labs/bridge-evm-contracts/pull/197/commits/b4a2487ce1e2e69ddb4bec9e7772515b7c06e30a) and [1e80398](https://github.com/bane-labs/bridge-evm-contracts/pull/197/commits/1e8039847f7bbef11b260ae1d221edbddc36eb6b).

- The `Impl` contracts in `contracts/bridge/` and `contracts/management/` (the ones in development) now hold the `reinitializer(_version)` functions (i.e., now `upgradeToV2()`) themselves.
- There are `Test...` contracts in `contracts/tests/` that extend the `Impl` contracts. They override some functions for testing reasons (e.g., functions with the `onlyAdmin` modifier), and hold some helper functions that are useful for tests but not needed in production.

In order to properly setup the contract's state in the test environment, so that we can test the reinitialization, the `Test...` contracts have an initialize function where the storage slot allocation will happen. I prefer to have this in these contracts instead of setting these storage allocations in the tests via hardhat/foundry to reduce duplication and additional complexity there. I added some comments inline in the `Test...` contract files for this.

So, now we have two contract versions and no V1, V2, ... confusion. One for production holding potential re-initialization logic, and another one that's just used for testing purposes.

contracts/management/BridgeManagementStorage.sol
mialbu: Fixed in 46493537051327003fc1c668fbfa2dfee8c58992

contracts/management/BridgeManagementStorageV1.sol
mialbu: This is a nice idea. I'll have a look at the EnumerableSet of OZ asap. If there's some issues that make it too complex, we can just use you suggestion with the position in the mapping instead of just a boolean. That'd be a good solution as well imo.

contracts/tests/TestBridgeManagement.sol
txhsl: LGTM.

contracts/management/BridgeManagementStorageV1.sol
mialbu: Can you elaborate on the problems to inherit the current storage? If you are referring to their warning on their docs to not delete a struct that contains a mapping, this should not be a problem. If we ever need to migrate the storage slots for that struct, we just need to make sure, we delete each entry in the set. 

If we use an EnumerableSet, we should migrate the validator array to that new set and delete the entries in the old array.

txhsl: > Can you elaborate on the problems to inherit the current storage?

> If we use an EnumerableSet, we should migrate the validator array to that new set and delete the entries in the old array.

Well, the problem is exactly the old `validators` array storage. If we are going to use OZ `EnumerableSet`, then we need to copy it to a new slot instead of reusing it. But if we only implement an `address=>position` mapping, then it can be reused.

> If we use an EnumerableSet, we should migrate the validator array to that new set and delete the entries in the old array.

So it is a trade-off, but I think both of them work.

mialbu: We need to add some migration logic anyway. I'll use the `EnumerableSet` because it reduces complexity and we can rely on audited logic for managing it instead of relying on custom code to add, remove, replace, etc.

mialbu: @txhsl I reworked the validators using OZ's EnumerableSet in commit [94580c0](https://github.com/bane-labs/bridge-evm-contracts/pull/197/commits/94580c054aa949d8f9e504e737b01f0cb768fc82). Please take a look.

contracts/management/BridgeManagementStorageV1.sol
```diff
@@ -27,6 +31,14 @@ abstract contract BridgeManagementStorageV1 is Ownable2StepUpgradeable {
...
-    // Slot 106
-    mapping(address validator => bool isValidator) internal validatorMap;
+    // Slot 106-107
+    EnumerableSet.AddressSet internal validatorSet;
+
+    function _upgradeToV2() internal onlyInitializing {
+        // Set validators in new version
+        uint256 validatorsLength = _v1_validators.length;
+        for (uint256 i = 0; i < validatorsLength; i++) {
+            EnumerableSet.add(validatorSet, _v1_validators[i]);
```
txhsl: It could be `validatorSet.add(_v1_validators[i]);`, as well as other operations.

And plz deal with the return value, it might be `false` if two same addresses are going to appear in the same set.

gsmachado: good catch.

mialbu: Oh, `validatorSet.add(_v1_validators[i]);` works? Let me check this. Maybe it was just an issue with the IDE I'm using.

mialbu: Fixed in [79a65e1](https://github.com/bane-labs/bridge-evm-contracts/pull/197/commits/79a65e1784aa22147cfcf3890d20eea0d1604ced)

txhsl (approved):

PR #204, opened 2024-10-29
Release 2.0.0

mialbu:
@txhsl Please compile the contracts and compare the artifacts to the ones present in `/release/2.0.0/`.

I used the existing `hardhat.config.ts` with:
- optimization runs: `200`
- solidity compiler: `0.8.25`
- evm version: `shanghai`


txhsl (approved):
Just to note, the compilation results are the same as in `/release/2.0.0/` when we import OpenZeppelin contracts under version `5.0.2`. But the current `package.json` imports them as `^5.0.2` and leads us to version `5.1.0`, then the result will be slightly different.

PR #214, opened 2025-01-29
Add Decimal Scaling Factor to Native Bridge Config

### LINKED ISSUE #208: [General EVM Chain Support] Extend native bridge data with `decimalScalingFactor`

mialbu:
Closes #208

### DIFF
--- a/contracts/bridge/BridgeStorageV1.sol
+++ b/contracts/bridge/BridgeStorageV1.sol
@@ -29,9 +29,34 @@ abstract contract BridgeStorageV1 is ReentrancyGuardUpgradeable {
     // Slot 104
     mapping(address tokenAddress => mapping(uint256 nonce => StorageTypes.Claimable claimable) claimableTokens) public
         tokenClaimables;
+
+    // Deprecated in V3 (slots will be cleared in V3 reinitialization)
     // Slots 105-113 (9 slots)
-    StorageTypes.NativeBridge public nativeBridge;
+    StorageTypes.NativeBridgeV2 public nativeBridgeV2;
 
     // Slot 114
     address[] public registeredTokens;
+
+    // Slot 115
+    StorageTypes.NativeBridgeV3 public nativeBridge;
+
+    function _upgradeToV3() internal onlyInitializing {
+        nativeBridge = StorageTypes.NativeBridgeV3(
+            nativeBridgeV2.paused,
+            nativeBridgeV2.depositState,
+            nativeBridgeV2.withdrawalState,
+            StorageTypes.NativeConfigV3(
+                nativeBridgeV2.config.fee,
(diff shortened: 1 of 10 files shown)


contracts/library/StorageTypes.sol
```diff
@@ -15,23 +15,38 @@ library StorageTypes {
...
+        uint256 maxAmount;
+        uint256 maxDeposits; // This should be used by the validators to decide for which deposit to sign if there are lots of deposits in a single block on the source chain, e.g., if this value is 50 and on the source chain there's 60 deposits in a single block, the resulting roots of deposit 50 and 60 should be signed and provided to the relayer.
+    }
+
+    struct NativeConfigV3 {
         uint256 fee;
         uint256 minAmount;
         uint256 maxAmount;
         uint256 maxDeposits; // This should be used by the validators to decide for which deposit to sign if there are lots of deposits in a single block on the source chain, e.g., if this value is 50 and on the source chain there's 60 deposits in a single block, the resulting roots of deposit 50 and 60 should be signed and provided to the relayer.
+        uint256 decimalScalingFactor;
```
txhsl, line 49: Only a new `uint256` field in the end, we may use the new struct directly, instead of creating a copy and having both V2 and V3 in the code.

It should return `0` if we try to access the old storage with `NativeConfigV3.decimalScalingFactor`, without any risky issue.

mialbu: No, the v2 struct uses as many slots as its size (right where it is). It uses the slots up to slot 103. If we just set the new one there, it will corrupt the storage. Decimal Scaling factor would access slot 104 which holds the size of the registeredTokens array, and the registered token array entries would no longer be accessible.

contracts/bridge/BridgeStorageV1.sol
```diff
@@ -29,9 +29,34 @@ abstract contract BridgeStorageV1 is ReentrancyGuardUpgradeable {
     // Slot 104
     mapping(address tokenAddress => mapping(uint256 nonce => StorageTypes.Claimable claimable) claimableTokens) public
         tokenClaimables;
+
+    // Deprecated in V3 (slots will be cleared in V3 reinitialization)
     // Slots 105-113 (9 slots)
-    StorageTypes.NativeBridge public nativeBridge;
+    StorageTypes.NativeBridgeV2 public nativeBridgeV2;
```
mialbu, line 35: Check the storage layout here, @txhsl. We can't just replace v2 with v3 because v3 would shift registeredTokens and use that slot for decimalScalingFactor.

contracts/library/StorageTypes.sol
txhsl: OK, I see.

txhsl (approved):

PR #215, opened 2025-01-29
feat: (#213) add function to set native bridge after deployment

### LINKED ISSUE #213: [General EVM Chain Support] Function to set native token bridge

mialbu:
Resolves #213

### DIFF
--- a/contracts/bridge/BridgeStorage.sol
+++ b/contracts/bridge/BridgeStorage.sol
@@ -57,6 +57,10 @@ abstract contract BridgeStorage is BridgeStorageV1, UUPSUpgradeable {
     error LengthMismatch();
     //0xa85293eb
     error MaxFeeExceeded(uint256 maxFeeAllowed, uint256 actualFee);
+    //0x5df7f28a
+    error NativeBridgeAlreadySet();
+    //0x2e56a1ef
+    error NativeBridgeNotSet();
     //0x79828e03
     error NoAuthorization();
     //0x475a97bd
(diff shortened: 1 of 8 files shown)


contracts/bridge/BridgeStorage.sol
```diff
@@ -120,6 +124,16 @@ abstract contract BridgeStorage is BridgeStorageV1, UUPSUpgradeable {
         _;
     }
 
+    modifier onlyIfNativeBridgeSet() {
```
txhsl, line 127: 🤔 Can we have `_setNativeBridge` in the constructor or upgrade method, and avoid using modifiers like this?

I mean, we always expect the native bridge is set when we call the contract, and we can't change the setting after we set it. So I think it's better to not check it every time in the entry, since it should be always `true`.

mialbu: Imagine we deploy on Ethereum. We might not support ETH right from the start, but still want to be able to use the bridge for other tokens.

txhsl: Maybe we initialize it but pause the native bridge?

txhsl (approved):
I'm not insisting, this works.

contracts/bridge/BridgeStorage.sol
mialbu: I see. Worth a thought, I agree. Let me take another look.

mialbu: I'll leave it as it is. A paused bridge means it exists and just cannot be used right now while an unset bridge is nonexistent. It's more concise that way especially if we're thinking about external integrations reading that information.

PR #221, opened 2025-06-17
Poc/arb msg

OT-rebel:


OT-rebel:
To execute just the 10 new tests run:
``` forge clean && forge test --match-test test_StoreAndExecuteMessage -vv --ffi```

File: contracts/bridge/BridgeImpl.sol [hunk 1]
```diff
@@ -581,4 +587,177 @@ contract BridgeImpl is BridgeStorage, IBridge, INativeBridge, ITokenBridge {
     function upgradeToV3() external virtual reinitializer(3) onlyAdmin {
         _upgradeToV3();
     }
+
+    // IMessageBridge Implementation
+
+    function storeMessage(uint nonce, bytes calldata message) public {
```
mialbu: issue: Right now, as from the executeMessage function these bytes are expected to be decodable into a `StorageTypes.Call` which does not have any metadata. We will need the message metadata here as well, i.e., the sender and time (a address and uint256). We can discuss if we want to embed that into the message bytes itself or separate.

contracts/library/StorageTypes.sol
```diff
@@ -68,4 +69,36 @@ library StorageTypes {
...
         uint256 decimalScalingFactor;
     }
+
+    // Message Bridge
+
+    struct Call {
+        address target;
+        bytes callData;
+        bool allowFailure;
+        uint256 value;
```
mialbu, line 79: thought: I'm not sure how we'll deal with how the value will be paid here. I think that will need to be paid entirely by the sender executing the message.

When talking to Guil yesterday, we thought of adding incentives to a call in the future. Once such functionality would be there, the message value here could be included in that as well.

contracts/bridge/BridgeImpl.sol
```diff
@@ -581,4 +587,177 @@ contract BridgeImpl is BridgeStorage, IBridge, INativeBridge, ITokenBridge {
...
+
+        // Verify that the msg.value matches the call.value from the message
+        if (msg.value != call.value) revert ValueMismatch(call.value, msg.value);
+
+        StorageTypes.Result memory result;
+
+        (result.success, result.returnData) = call.target.call{value: call.value}(call.callData);
+
+        // forward the reason for failure if the call was not allowed to fail
+        if (!call.allowFailure && !result.success) revert CallFailed(result.returnData);
```
mialbu, line 605: suggestion: maybe we can invert the `allowFailure` boolean and call it `requireSuccess` to make this logic here a bit more intuitive.

contracts/library/StorageTypes.sol
OT-rebel: Yeah, this needs to be talked about. currently the executor is required to pay for the value.
I can take it out or we can mandate that it's always 0. I initially added it just to test it and see the limitations

contracts/bridge/BridgeImpl.sol
```diff
@@ -581,4 +587,177 @@ contract BridgeImpl is BridgeStorage, IBridge, INativeBridge, ITokenBridge {
...
+    function executeMessage(uint nonce) public payable returns (StorageTypes.Result memory) {
+        if (n3ToEvmMessages[nonce].length == 0) revert MessageNotFound(nonce);
+        StorageTypes.Call memory call = abi.decode(n3ToEvmMessages[nonce], (StorageTypes.Call));
+
+        // Verify that the msg.value matches the call.value from the message
+        if (msg.value != call.value) revert ValueMismatch(call.value, msg.value);
+
+        StorageTypes.Result memory result;
+
+        (result.success, result.returnData) = call.target.call{value: call.value}(call.callData);
```
mialbu, line 602: I know you're also aware of this, but just to keep track of this I added this comment nonetheless.

issue:  this is dangerous as it could allow various actions to be performed on behalf of the bridge contract, e.g., one could drain the contract's entire ERC20 balances by using a call that approves the entire balance to any spender. Please add a todo here that we must forward this call to be executed in the context of a separate new contract other than this bridge contract.

Reference: See OZ's implementation of the `approve` function of the ERC20 standard: https://github.com/OpenZeppelin/openzeppelin-contracts/blob/668221e314409ec3a32965b3a66afa7fc0b64b33/contracts/token/ERC20/ERC20.sol#L120-L124

contracts/library/MessageBridgeLib.sol
```diff
@@ -0,0 +1,32 @@
...
+    function _computeNewTopRoot(
+        bytes32 _previousRoot,
+        StorageTypes.MessageData[] memory _messages
+    ) internal pure returns (bytes32) {
+        bytes32 parent = _previousRoot;
+        uint256 messagesLength = _messages.length;
+
+        for (uint256 i = 0; i < messagesLength; i++) {
+            StorageTypes.MessageData memory messageData = _messages[i];
+            bytes32 messageHash = keccak256(abi.encodePacked(messageData.nonce, messageData.message));
```
mialbu, line 26: todo: please add a separate function for this step. I'd like that we can test this and sync with the replicate part on N3.

For example: `hashMessage(StorageTypes.MessageData memory _message) internal pure returns (bytes32)`.

PR #242, opened 2025-07-02
Integrate Metadata into Messages

### LINKED ISSUE #225: [AMB] [Store] Include metadata

OT-rebel:
Closes #225 

- Add metadata to `Message` struct
- Store full `Message` structs instead of decoded call bytes - call bytes no longer get decoded upon storing.


contracts/bridge/BridgeImpl.sol
```diff
@@ -724,9 +724,22 @@ contract BridgeImpl is BridgeStorage, IBridge, INativeBridge, ITokenBridge, IMes
         // Store each message
         for (uint256 i = 0; i < messageLength; i++) {
             StorageTypes.MessageData calldata messageData = _messages[i];
-            if (n3ToEvmMessages[messageData.nonce].target != address(0)) revert MessageAlreadyExists(messageData.nonce);
-            // Decode the message bytes into a Call struct and store it directly
-            n3ToEvmMessages[messageData.nonce] = abi.decode(messageData.message, (StorageTypes.Call));
+            // Verify message format by decoding it (will revert if not a valid Call structure)
+            StorageTypes.Call memory call = abi.decode(messageData.message, (StorageTypes.Call));
```
mialbu: issue: This is a critical part. I'm not sure we can (and even if we should) verify on N3 that the provided bytes can be decoded to a StorageTypes.Call. Imo, we should leave that responsibility to the users and not the bridge.

I don't see the point in decoding at this point. We can just store the bytes and be done with it. The responsibility of the bridge should be moving data, i.e., bytes. It should not be responsible for validating the format of these bytes.

contracts/bridge/BridgeImpl.sol
```diff
@@ -724,9 +724,22 @@ contract BridgeImpl is BridgeStorage, IBridge, INativeBridge, ITokenBridge, IMes
...
+
+            // Verify this message hasn't been stored before
+            // We still check using the target field from the decoded Call as a proxy for existence
+            if (call.target == address(0)) revert InvalidCallTarget();
+
+            // Check if a message with this nonce already exists by checking if there's raw message data
+            if (n3ToEvmRawMessages[messageData.nonce].length > 0) revert MessageAlreadyExists(messageData.nonce);
```
mialbu: issue: this should be dead code if the hash-chain checks with the subsequent nonce check above. I'm not sure it should be here at all.

contracts/bridge/BridgeImpl.sol
```diff
@@ -591,8 +591,8 @@ contract BridgeImpl is BridgeStorage, IBridge, INativeBridge, ITokenBridge, IMes
     // IMessageBridge Implementation
 
     function executeMessage(uint nonce) public payable returns (StorageTypes.Result memory) {
-        StorageTypes.Call memory call = n3ToEvmMessages[nonce];
-        if (call.target == address(0)) revert MessageNotFound(nonce);
+        if (n3ToEvmRawMessages[nonce].length == 0) revert MessageNotFound(nonce);
+        StorageTypes.Call memory call = abi.decode(n3ToEvmRawMessages[nonce], (StorageTypes.Call));
```
mialbu: question: is it cheaper to read the storage entry twice rather than load the bytes to memory and assign it a separate variable? If the difference in cost is not huge, I'd prefer to assign `n3ToEvmRawMessages[none]` to a variable.

contracts/bridge/BridgeImpl.sol
```diff
@@ -724,9 +724,22 @@ contract BridgeImpl is BridgeStorage, IBridge, INativeBridge, ITokenBridge, IMes
...
+            if (call.target == address(0)) revert InvalidCallTarget();
+
+            // Check if a message with this nonce already exists by checking if there's raw message data
+            if (n3ToEvmRawMessages[messageData.nonce].length > 0) revert MessageAlreadyExists(messageData.nonce);
+
+            // Store the raw message bytes
+            n3ToEvmRawMessages[messageData.nonce] = messageData.message;
+
+            // Store metadata separately
+            n3ToEvmMetadata[messageData.nonce] = messageData.metadata;
```
mialbu: task: please elaborate on the reason to use two mappings storing the raw message and the metadata separately.

mialbu (changes requested):

contracts/bridge/BridgeImpl.sol
OT-rebel: yeah this was leftover from when I was storing `Call`s instead of raw bytes. I didn't revert properly

contracts/bridge/BridgeImpl.sol
OT-rebel: I guess it's just a force of habit to separate items in different mappings, but these do actually belong together since they won't be re-written again

contracts/bridge/BridgeImpl.sol
OT-rebel: no, it's not cheaper at all. But this changes a bit if I store metadata and rawMessage together so this will be implicitly solved

contracts/bridge/BridgeImpl.sol

contracts/bridge/BridgeImpl.sol

contracts/bridge/BridgeImpl.sol
```diff
@@ -693,19 +693,27 @@ contract BridgeImpl is BridgeStorage, IBridge, INativeBridge, ITokenBridge, IMes
...
+
+        emit MessageDeposit(messageData.nonce, messageData.message);
+    }
+
     function executeMessage(uint256 nonce) external payable returns (StorageTypes.Result memory) {
-        StorageTypes.Call memory call = n3ToEvmMessages[nonce];
-        if (call.target == address(0)) revert MessageNotFound(nonce);
+        bytes memory storedMessage = n3ToEvmMessages[nonce].message;
+        if (storedMessage.length == 0) revert MessageNotFound(nonce);
+        StorageTypes.Call memory call = abi.decode(storedMessage, (StorageTypes.Call));
```
mialbu, line 717: todo: add a todo here that the decoding should be done in the execution contract - just preparing for awareness in #243.

test/MessageBridge.t.sol
```diff
@@ -542,12 +551,11 @@ contract MessageBridgeTest is Test, SigUtils {
...
 
         // Store the message with the nonce
-        storeMessage(nonce, message);
+        storeMessage(nonce, message, "");
 
         // Try to store the same message with the same nonce again: should revert with MessageAlreadyExists
-        vm.expectRevert(abi.encodeWithSelector(BridgeStorage.MessageAlreadyExists.selector, nonce));
         vm.prank(relayer);
-        storeMessage(nonce, message);
+        storeMessage(nonce, message, BridgeStorage.InvalidNonceSequence.selector);
```
mialbu: issue: the selector seems to be a wrong type for this function.

Use `abi.encodeWithSelector(BridgeStorage.InvalidNonceSequence.selector)` instead.

contracts/library/StorageTypes.sol
```diff
@@ -85,6 +72,32 @@ library StorageTypes {
...
     struct MessageData {
         uint256 nonce;
+        Metadata metadata;
+        bytes message;
+    }
+
+    struct Metadata {
+        // Pack 32+64+160=256 bits into a single 32-byte slot
+        uint32 version; // For forward compatibility, should be incremented if the structure changes
+        uint64 chainId;
```
mialbu: issue: we don't need the chain id here. Any message sent in the context of a deployed bridge contract will only come from a distinct N3 chain.

contracts/bridge/BridgeStorage.sol
```diff
@@ -97,6 +93,10 @@ abstract contract BridgeStorage is BridgeStorageV1, UUPSUpgradeable {
     error InvalidMessageSize();
     //0x000bf7e9
     error MessageRootMismatch();
+    //0x05eae184
+    error InvalidCallTarget();
```
mialbu: issue: This error is unused. Please remove it.

mialbu (changes requested):

contracts/bridge/BridgeImpl.sol
OT-rebel: done

test/MessageBridge.t.sol
OT-rebel: done

contracts/library/StorageTypes.sol
OT-rebel: done

contracts/bridge/BridgeStorage.sol
OT-rebel: done

mialbu (approved):

PR #243, opened 2025-07-04
Add separate ExecutionManager to handle arb. message execution

### LINKED ISSUE #237: Use an Executor Contract

OT-rebel:
Closes #237


contracts/bridge/BridgeImpl.sol
```diff
@@ -710,21 +719,29 @@ contract BridgeImpl is BridgeStorage, IBridge, INativeBridge, ITokenBridge, IMes
         emit MessageDeposit(messageData.nonce, messageData.message);
     }
 
-    function executeMessage(uint256 nonce) external payable returns (StorageTypes.Result memory) {
+    function executeMessage(
+        uint256 nonce,
+        address delegatedExecutor
+    )
+        public
```
mialbu:
```suggestion
        external
```

contracts/bridge/BridgeImpl.sol
```diff
@@ -710,21 +719,29 @@ contract BridgeImpl is BridgeStorage, IBridge, INativeBridge, ITokenBridge, IMes
...
         if (storedMessage.length == 0) revert MessageNotFound(nonce);
-        // TODO: decode this in the executor
-        StorageTypes.Call memory call = abi.decode(storedMessage, (StorageTypes.Call));
 
-        // Verify that the msg.value matches the call.value from the message
-        if (msg.value != call.value) revert ValueMismatch(call.value, msg.value);
+        if (address(executionManager) == address(0)) revert ExecutionManagerNotSet();
 
-        StorageTypes.Result memory result;
+        // Mark as executed
```
mialbu, line 730: todo: this is a critical step here. I'm noting it here that we ensure this is appropriately implemented before approval and merging.

contracts/bridge/ExecutionManager.sol
```diff
@@ -0,0 +1,69 @@
...
+contract ExecutionManager is IExecutionManager, AccessControl {
+    bytes32 public constant BRIDGE_ROLE = keccak256("BRIDGE_ROLE");
+
+    error UnexpectedDelegation();
+    error UnableToExecuteCall();
+    //0x626ade30
+    error ValueMismatch(uint256 providedValue, uint256 expectedValue);
+
+    constructor(address admin, address bridge) {
+        _grantRole(DEFAULT_ADMIN_ROLE, admin);
```
mialbu: question: do we need an admin?

contracts/bridge/ExecutionManager.sol
```diff
@@ -0,0 +1,69 @@
...
+        address delegatedExecutor
+    )
+        external
+        payable
+        override
+        onlyRole(BRIDGE_ROLE)
+        returns (StorageTypes.Result memory)
+    {
+        // Decode the raw message into a StorageTypes.Call struct
+        StorageTypes.Call memory call = abi.decode(rawMessage, (StorageTypes.Call));
```
mialbu, line 33: documentation: please add a comment about what happens in case the `rawMessage` cannot be decoded to a `StorageTypes.Call` struct.

contracts/bridge/BridgeImpl.sol
```diff
@@ -710,21 +719,29 @@ contract BridgeImpl is BridgeStorage, IBridge, INativeBridge, ITokenBridge, IMes
         emit MessageDeposit(messageData.nonce, messageData.message);
     }
 
-    function executeMessage(uint256 nonce) external payable returns (StorageTypes.Result memory) {
+    function executeMessage(
+        uint256 nonce,
+        address delegatedExecutor
```
mialbu: issue: An execution delegation should be set when sending the message on N3. I don't think we need this parameter here.

contracts/bridge/ExecutionManager.sol
```diff
@@ -0,0 +1,69 @@
...
+
+        bool success;
+        bytes memory returnData;
+
+        // Check that this is the executing contract
+        if (call.executingContract == address(this)) {
```
mialbu: nitpick: Imo, the `executingContract` needs renaming. The term "executing" sounds like a factual value like this actually is the executing contract instead of what it should be.

suggestion: executionContract, delegatedExecutor or even delegatedExecutionContract

contracts/bridge/ExecutionManager.sol
```diff
@@ -0,0 +1,69 @@
...
+            if (delegatedExecutor != address(0)) {
+                revert UnexpectedDelegation();
+            } else {
+                (success, returnData) = _executeCall(call.target, call.callData, call.value);
+
+                if (!success && !call.allowFailure) revert(string(returnData));
```
mialbu: issue: I'm not sure casting the returnData to a string is optimal, at least not if we just revert with only that value. If we use the returnData in some way, I think it's better to use a custom error here to make it more concise indicating that the call is not allowed to fail but it failed.

contracts/bridge/ExecutionManager.sol
```diff
@@ -0,0 +1,69 @@
...
+        address target,
+        bytes memory callData,
+        uint256 value
+    )
+        internal
+        returns (bool success, bytes memory returnData)
+    {
+        if (msg.value < value) revert ValueMismatch(msg.value, value);
+
+        (success, returnData) = target.call{value: value}(callData);
```
mialbu: thought: is there any edge case here that could be misused, i.e., for which we should limit the `target` value?

contracts/bridge/ExecutionManager.sol
```diff
@@ -0,0 +1,69 @@
...
+                revert UnexpectedDelegation();
+            } else {
+                (success, returnData) = _executeCall(call.target, call.callData, call.value);
+
+                if (!success && !call.allowFailure) revert(string(returnData));
+            }
+        } else if (delegatedExecutor == address(0)) {
+            revert UnableToExecuteCall();
+        } else {
+            (success, returnData) = IDelegatedExecutor(delegatedExecutor).executeCall{value: msg.value}(
```
mialbu: thought: is there any edge case here that could be misused, i.e., for which we should limit the `delegatedExecutor` value?

mialbu (changes requested):

contracts/bridge/ExecutionManager.sol
OT-rebel: I think that we should exactly forward `msg.value` instead so this way there's no chance that any funds remain or are taken from this contract other than what comes in

contracts/bridge/ExecutionManager.sol
OT-rebel: not really. this was required before, when this was configurable with an allowlist. I'll remove

contracts/bridge/ExecutionManager.sol
OT-rebel: none that I can see since we only pass along the `msg.value` meaning that if someone calls this via re-entrancy it will have to provide the value as well

OT-rebel: I could make a test for re-entrancy with this type of scenario

contracts/bridge/BridgeImpl.sol

contracts/bridge/BridgeImpl.sol
OT-rebel: done

contracts/bridge/ExecutionManager.sol
OT-rebel: done

contracts/bridge/ExecutionManager.sol
OT-rebel: done

contracts/bridge/BridgeImpl.sol

contracts/bridge/ExecutionManager.sol
OT-rebel: done

contracts/bridge/ExecutionManager.sol
OT-rebel: done

contracts/bridge/ExecutionManager.sol
OT-rebel: done

contracts/bridge/ExecutionManager.sol
OT-rebel: Added issue #250 for re-entrancy test

contracts/library/StorageTypes.sol
```diff
@@ -85,17 +85,21 @@ library StorageTypes {
...
+        bool requiresResponse; // The user can specify if the call requires a response to be sent back across the bridge.
+        address target;
+        address delegatedExecutor; // This is the contract that will call the `target` with the `callData`. 0x00..00 implies that the message executor will call the target directly.
         uint256 value;
+        bytes callData;
     }
 
     struct Result {
         bool success;
+        bool requiresResponse; // Indicates if the result should be sent back across the bridge.
```
mialbu: thought: imo, this is something that could be decided when sending the message. That means, it's up to whoever sends it, not dependent on what's the result of the call.

I see three main cases we should cover once we get to the result part:
1. the sender doesn't need a result
2. the sender doesn't need a result right away
3. the sender needs a result immediately

In all cases, I'd use an event to emit the result. In case 2, we'd store the result (to be bridged back in the future if desired), while in case 3 we could initiate an immediate return of the result within the scope of the bridge's executeMessage function.

Having said all that, I don't think we need this boolean here as it should be decided in the very beginning of the message's journey.

contracts/library/StorageTypes.sol
```diff
@@ -85,17 +85,21 @@ library StorageTypes {
...
+        bool executed;
     }
 
     struct Call {
-        address target;
-        bytes callData;
         bool allowFailure;
+        bool requiresResponse; // The user can specify if the call requires a response to be sent back across the bridge.
+        address target;
+        address delegatedExecutor; // This is the contract that will call the `target` with the `callData`. 0x00..00 implies that the message executor will call the target directly.
```
mialbu: Imo, we can remove the concept of a delegated executor for now and just stick with using the execution manager contract we're building. In a future step we could introduce a delegation feature if that's something users are asking for. Please remove the related logic in the execution manager accordingly.

contracts/bridge/BridgeImpl.sol
```diff
@@ -1,18 +1,21 @@
...
+import {BridgeLib} from "../library/BridgeLib.sol";
+import {MessageBridgeLib} from "../library/MessageBridgeLib.sol";
+import {NativeBridgeLib} from "../library/NativeBridgeLib.sol";
+import {StorageTypes} from "../library/StorageTypes.sol";
+import {TokenBridgeLib} from "../library/TokenBridgeLib.sol";
+import {BridgeStorage} from "./BridgeStorage.sol";
+import {BridgeStorageV1} from "./BridgeStorageV1.sol";
+import {ReentrancyGuardUpgradeable} from "@openzeppelin/contracts-upgradeable/utils/ReentrancyGuardUpgradeable.sol";
+import {IERC20} from "@openzeppelin/contracts/token/ERC20/IERC20.sol";
+import {SafeERC20} from "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
```
mialbu, line 18: I like these explicit imports! We should use them in all our contracts. I've created an issue for it: #251.

contracts/bridge/BridgeImpl.sol
```diff
@@ -590,6 +593,12 @@ contract BridgeImpl is BridgeStorage, IBridge, INativeBridge, ITokenBridge, IMes
 
     // IMessageBridge Implementation
 
+    // Function to set the message executor
+    function setMessageExecutor(address _executor) external override onlyGovernor {
+        executionManager = IExecutionManager(_executor);
+        emit MessageExecutorSet(_executor);
+    }
```
mialbu: thought: what do you think about moving the message execution manager address to the message bridge config?

contracts/bridge/BridgeImpl.sol
mialbu: critical: We also need a check that the message hasn't been executed already in the beginning of the `executeMessage()` function. Currently, the `executed` boolean value has no effect.

mialbu (changes requested):

contracts/bridge/ExecutionManager.sol
```diff
@@ -0,0 +1,73 @@
...
+                nonce, call.target, call.callData, call.value
+            );
+        }
+        return StorageTypes.Result({success: success, requiresResponse: call.requiresResponse, returnData: returnData});
+    }
+
+    function _executeCall(
+        address target,
+        bytes memory callData,
+        uint256 value
```
mialbu: nitpick: order these consistently. Either change the order in the `Call` struct (move the `value` after `callData`) or reverse the parameter order here.

contracts/library/StorageTypes.sol
OT-rebel: well, this is a bit tricky because this Response is coming out of the ExecutionManager since the Bridge itself does not know if a message requires the result to be sent back across the bridge. 
Though maybe I could just have the ExecutionManager return a tuple and not have this be part of the Result struct

contracts/bridge/BridgeImpl.sol
OT-rebel: actually this was trivial and I kinda already did it

contracts/bridge/BridgeImpl.sol
OT-rebel: To me it feels closer to what the `IBridgeManagement` is than what a `MessageConfig` component is.

I mean, sure, we made it for the `MessageBridge`, but we could even use it for the ERC20 transfers that are currently not allowed in the bridge

contracts/library/StorageTypes.sol
OT-rebel: removed

contracts/bridge/ExecutionManager.sol
OT-rebel: done

contracts/library/StorageTypes.sol
mialbu: Imo, this doesn't need to be part of the serialized `Call`. It would be part of the metadata - a configurable part of it.

That way the executionManager is only responsible for executing and returning the result in its `executeMessage()` function and the bridge would be responsible for further action.

contracts/bridge/BridgeImpl.sol
mialbu: > use it for the ERC20 transfers that are currently not allowed in the bridge

What do you mean?

OT-rebel: I mean the ERC20 transfers that are made to a contract and they are not currently executed but someone has to come and execute those

contracts/bridge/ExecutionManager.sol
```diff
@@ -0,0 +1,53 @@
...
+        address target,
+        uint256 value,
+        bytes memory callData
+    )
+        private
+        returns (bool success, bytes memory returnData)
+    {
+        if (msg.value < value) revert ValueMismatch(msg.value, value);
+
+        (success, returnData) = target.call{value: msg.value}(callData);
```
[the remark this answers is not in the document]
OT-rebel: will enhance in #250 to make sure that the re-entrancy is not usable

contracts/library/StorageTypes.sol
OT-rebel: sending the response back across the bridge is not very standard afaik and since I expect that the response will require a supplementary fee, I think users should be the ones to decide if they want to receive an answer, regardless of what the result of the message execution is

contracts/bridge/BridgeImpl.sol
OT-rebel: but that's just a thought I had about how we could use the ExecutionManager even outside the MessageBridge, not necessarily something we should do

contracts/bridge/BridgeImpl.sol
OT-rebel: yeah, somehow I missed this earlier

OT-rebel: it's done now

contracts/library/StorageTypes.sol
mialbu: Let's continue this once we take on #247. For now, let's just emit the result in an event and store it in the bridge contract.

Summary of our internal discussion: We could tag messages to discriminate what type/kind of message it is. Initially using 2 tags declaring whether the message:
1. should be stored to the executable mapping (the bytes should be an encoded Call that can be executed), or 
2. should be stored to the result mapping -> the bytes are result bytes from N3

This way, the result of the execution of a message sent to N3 could utilize the message bridge to send back the result bytes.

mialbu (approved):

PR #244, opened 2025-07-07
Arb msg Base code

OT-rebel:


contracts/bridge/BridgeStorageV1.sol
```diff
@@ -40,6 +40,12 @@ abstract contract BridgeStorageV1 is ReentrancyGuardUpgradeable {
     // Slot 115
     StorageTypes.NativeBridgeV3 public nativeBridge;
 
+    // Slot 116
+    StorageTypes.MessageBridge public messageBridge;
+
+    // Slot 117
+    mapping(uint256 => StorageTypes.Call) public n3ToEvmMessages;
+
```
mialbu: issue: no code change needed, but the comments need to be corrected

explanation:
Since we use the storage layout based on the definition sequence here in the solidity files, we need to be aware of which slot exactly is used for what value. Here's a quick explanation:

`nativeBridge` consumes 10 slots. `messageBridge` will start at slot 125 and consume 8 slots (125-132). Thus, then `n3ToEvmMessages` will be in slot 133 (and just use 1 slot).

You can use `forge inspect contracts/management/BridgeImpl.sol:BridgeImpl storageLayout` to get the storage layout.

```solidity
struct MessageBridge {
    bool paused; // slot 125 (a full slot because the next value (nonce of n3ToEvmState) is a uint256 which needs a full slot itself)
    State n3ToEvmState; // slots 126,127
    State evmToN3State; // slots 128,129
    MessageConfig config; // slots 130,131,132
}

// 2 slots
struct State {
    uint256 nonce;
    bytes32 root;
}

// 3 slots
struct MessageConfig {
    uint256 fee;
    uint256 maxMessageSize;
    uint256 maxDeposits;
}

File: contracts/bridge/BridgeImpl.sol [hunk 2]
```diff
@@ -581,4 +587,174 @@ contract BridgeImpl is BridgeStorage, IBridge, INativeBridge, ITokenBridge {
...
+
+        StorageTypes.Result memory result;
+
+        (result.success, result.returnData) = call.target.call{value: call.value}(call.callData);
+
+        // forward the reason for failure if the call was not allowed to fail
+        if (!call.allowFailure && !result.success) revert CallFailed(result.returnData);
+
+        return result;
+    }
```
mialbu: nitpick: move this function after `storeMessage()`. If there's a clear flow associated it's easier to read a contract if the functions are also ordered appropriately, i.e., `executeMessage()` after `storeMessage()`.

contracts/bridge/BridgeImpl.sol
```diff
@@ -581,4 +587,174 @@ contract BridgeImpl is BridgeStorage, IBridge, INativeBridge, ITokenBridge {
...
+     */
+    function messageBridgeIsSet() external view override returns (bool) {
+        return _messageBridgeIsSet();
+    }
+
+    /**
+     * @notice Set up the message bridge with configuration parameters.
+     * @param _fee the fee for using the message bridge.
+     * @param _maxMessageSize the maximum allowed size of a message in bytes.
+     * @param _maxDeposits the maximum number of deposits that can be processed in a single transaction.
```
mialbu:
```suggestion
     * @param _maxNrMessages the maximum number of messages that can be processed in a single transaction by the relayer.
```

contracts/bridge/BridgeImpl.sol
```diff
@@ -581,4 +587,174 @@ contract BridgeImpl is BridgeStorage, IBridge, INativeBridge, ITokenBridge {
...
+     */
+    function setMessageBridge(
+        uint256 _fee,
+        uint256 _maxMessageSize,
+        uint256 _maxDeposits
```
mialbu: suggestion: use `_maxNrMessages` instead of `_maxDeposits`.

contracts/bridge/BridgeStorage.sol
```diff
@@ -394,6 +428,76 @@ abstract contract BridgeStorage is BridgeStorageV1, UUPSUpgradeable {
...
+    // Message Bridge functions
+
+    function _messageBridgeIsSet() internal view returns (bool) {
+        return messageBridge.config.maxMessageSize != 0;
+    }
+
+    function _setMessageBridge(
+        uint256 _fee,
+        uint256 _maxMessageSize,
+        uint256 _maxDeposits
```
mialbu: suggestion: use `_maxNrMessages` instead of `_maxDeposits`.

contracts/bridge/BridgeImpl.sol
```diff
@@ -581,4 +587,174 @@ contract BridgeImpl is BridgeStorage, IBridge, INativeBridge, ITokenBridge {
...
+
+        // Check if nonces are in sequence
+        uint256 lastNonce = state.nonce;
+        for (uint256 i = 0; i < messageLength; i++) {
+            if (_messages[i].nonce != lastNonce + 1) revert InvalidNonceSequence();
+            lastNonce = _messages[i].nonce;
+
+            // Check message size
+            if (_messages[i].message.length > config.maxMessageSize) revert InvalidMessageSize();
+        }
```
mialbu: suggestion: There might be a cheaper way to achieve the same by not needing to update a variable's value in memory. Take a look at how we've checked the sequence in `DepositData` arrays here: https://github.com/bane-labs/bridge-evm-contracts/blob/186ce2908bf6765eb74465732a63f7ec97ccb528/contracts/library/BridgeLib.sol#L18-L24

contracts/library/MessageBridgeLib.sol
```diff
@@ -0,0 +1,32 @@
...
+    function _computeNewTopRoot(
+        bytes32 _previousRoot,
+        StorageTypes.MessageData[] memory _messages
+    ) internal pure returns (bytes32) {
+        bytes32 parent = _previousRoot;
+        uint256 messagesLength = _messages.length;
+
+        for (uint256 i = 0; i < messagesLength; i++) {
+            StorageTypes.MessageData memory messageData = _messages[i];
+            bytes32 messageHash = keccak256(abi.encodePacked(messageData.nonce, messageData.message));
```
mialbu: issue: please move the creation of this individual `messageHash` in a separate function, i.e., `_hashMessageBridgeOp()`.

contracts/bridge/BridgeImpl.sol
```diff
@@ -581,4 +587,174 @@ contract BridgeImpl is BridgeStorage, IBridge, INativeBridge, ITokenBridge {
...
+        emit MessageDepositRootUpdate(_messages[messageLength - 1].nonce, _depositRoot);
+
+        // Store each message
+        for (uint256 i = 0; i < messageLength; i++) {
+            StorageTypes.MessageData calldata messageData = _messages[i];
+            if (n3ToEvmMessages[messageData.nonce].target != address(0)) revert MessageAlreadyExists(messageData.nonce);
+            // Decode the message bytes into a Call struct and store it directly
+            n3ToEvmMessages[messageData.nonce] = abi.decode(messageData.message, (StorageTypes.Call));
+            emit MessageDeposit(messageData.nonce, messageData.message);
+        }
```
mialbu, line 703: issue: please add this to a separate private function (or add a todo for now at least - same for the subsequent check code above).

mialbu (changes requested):

contracts/bridge/BridgeImpl.sol
OT-rebel: There's changes made to this in the metadata branch. we can revisit when reviewing that

contracts/bridge/BridgeImpl.sol
```diff
@@ -581,4 +587,175 @@ contract BridgeImpl is BridgeStorage, IBridge, INativeBridge, ITokenBridge {
...
+    function setMaxMessageSize(uint256 _maxSize) external override onlyGovernor {
+        _setMaxMessageSize(_maxSize);
+        emit MaxMessageSizeChange(_maxSize);
+    }
+
+    /**
+     * @notice Set the maximum number of messages that can be processed in a single transaction.
+     * @param _maxNrMessages the new maximum number of messages.
+     */
+    function setMaxNrMessages(uint256 _maxNrMessages) external override onlyGovernor {
```
mialbu, line 745: issue: add this function to the IMessageBridge interface as it currently doesn't override anything. I couldn't compile the contracts because of this.

contracts/bridge/BridgeStorage.sol
```diff
@@ -394,6 +428,76 @@ abstract contract BridgeStorage is BridgeStorageV1, UUPSUpgradeable {
...
+        if (_maxDeposits == 0) revert InvalidValue();
+
+        messageBridge = StorageTypes.MessageBridge({
+            paused: true,
+            n3ToEvmState: StorageTypes.State({nonce: 0, root: 0x0}),
+            evmToN3State: StorageTypes.State({nonce: 0, root: 0x0}),
+            config: StorageTypes.MessageConfig({
+                fee: _fee,
+                maxMessageSize: _maxMessageSize,
+                maxDeposits: _maxDeposits
```
mialbu: issue: needs renaming to match the new term `maxNrMessages`.

contracts/bridge/BridgeStorage.sol
```diff
@@ -394,6 +428,76 @@ abstract contract BridgeStorage is BridgeStorageV1, UUPSUpgradeable {
...
+    }
+
+    function _setMaxMessageSize(uint256 _maxMessageSize) internal {
+        if (_maxMessageSize == 0) revert InvalidValue();
+        messageBridge.config.maxMessageSize = _maxMessageSize;
+    }
+
+    function _setMaxMessageDeposits(uint256 _maxDeposits) internal {
+        if (_maxDeposits == 0) revert InvalidValue();
+        messageBridge.config.maxDeposits = _maxDeposits;
```
mialbu: issue: needs renaming to match the new term `maxNrMessages`.

contracts/bridge/BridgeImpl.sol
```diff
@@ -581,4 +587,175 @@ contract BridgeImpl is BridgeStorage, IBridge, INativeBridge, ITokenBridge {
...
+        if (messageLength > config.maxNrMessages) revert InvalidDepositsLength();
+
+        // Check if nonces are in sequence
+        // More gas-efficient nonce validation that doesn't update a variable on each iteration
+        // Each nonce should be exactly (state.nonce + position in array + 1)
+        for (uint256 i = 0; i < messageLength; i++) {
+            if (_messages[i].nonce != state.nonce + i + 1) revert InvalidNonceSequence();
+
+            // Check message size
+            if (_messages[i].message.length > config.maxMessageSize) revert InvalidMessageSize();
```
mialbu: issue: We should omit this check.

In my opinion this is something that should be enforced on the originating and not on the destination chain. Think about it this way: If some configuration on N3 is misaligned for some reason, this could allow larger incoming messages on N3 and the bridging would be halted as the relayer would not be able to create a non-reverting transaction invoking this function with that data.

It's somewhat similar to the `maxAmount` value in the token bridges. There this cap is enforced on the origin chain, but on the destination chain it's not checked or enforced again.

contracts/bridge/BridgeImpl.sol
```diff
@@ -581,4 +587,175 @@ contract BridgeImpl is BridgeStorage, IBridge, INativeBridge, ITokenBridge {
...
+        // Update the message bridge deposit state
+        _setMessageBridgeN3ToEvmState(
+            StorageTypes.State({nonce: _messages[messageLength - 1].nonce, root: _depositRoot})
+        );
+        emit MessageDepositRootUpdate(_messages[messageLength - 1].nonce, _depositRoot);
+
+        // TODO: extract to a private function
+        for (uint256 i = 0; i < messageLength; i++) {
+            StorageTypes.MessageData calldata messageData = _messages[i];
+            if (n3ToEvmMessages[messageData.nonce].target != address(0)) revert MessageAlreadyExists(messageData.nonce);
```
mialbu, line 699: issue: if this would be triggered, something would be wrong in the logic elsewhere, thus, I'm inclined in using assert here or omitting it entirely. We don't have such a check in `_addTokenClaimable()` either.

contracts/bridge/BridgeStorage.sol
mialbu: please rename this.

contracts/bridge/BridgeImpl.sol
mialbu: done, storeMessage is now above `executeMessage`.

contracts/bridge/BridgeImpl.sol
OT-rebel: added a TODO above this codeblock

contracts/bridge/BridgeImpl.sol
```diff
@@ -581,4 +587,175 @@ contract BridgeImpl is BridgeStorage, IBridge, INativeBridge, ITokenBridge {
...
+            // Decode the message bytes into a Call struct and store it directly
+            n3ToEvmMessages[messageData.nonce] = abi.decode(messageData.message, (StorageTypes.Call));
```
mialbu, line 701: todo: please add a todo here to not force `messageData.message` into a predefined struct, instead it should just be a byte array.

contracts/bridge/BridgeImpl.sol
```diff
@@ -581,4 +587,175 @@ contract BridgeImpl is BridgeStorage, IBridge, INativeBridge, ITokenBridge {
...
+
+        // Verify that the msg.value matches the call.value from the message
+        if (msg.value != call.value) revert ValueMismatch(call.value, msg.value);
+
+        StorageTypes.Result memory result;
+
+        (result.success, result.returnData) = call.target.call{value: call.value}(call.callData);
+
+        // forward the reason for failure if the call was not allowed to fail
+        if (!call.allowFailure && !result.success) revert CallFailed(result.returnData);
```
mialbu, line 718: issue: I'm not certain if we should to revert here. We might just want to return the result value.

I need to think a bit more about this one. Please add a todo to keep this in consideration.

We'll certainly come back to this when working on #247.

mialbu (changes requested):

contracts/bridge/BridgeImpl.sol
OT-rebel: I think this was entirely removed in the metadata as we agreed to store the message as they are, not as `Call`s so I'm pretty sure this was removed

contracts/bridge/BridgeImpl.sol
OT-rebel: same as the other line above

contracts/bridge/BridgeImpl.sol
OT-rebel: I added a comment in that issue rather than in the code and we'll tackle it at that time

contracts/bridge/BridgeImpl.sol
OT-rebel: done

contracts/bridge/BridgeStorage.sol
OT-rebel: done

contracts/bridge/BridgeStorage.sol
OT-rebel: done

contracts/bridge/BridgeImpl.sol

contracts/bridge/BridgeImpl.sol
mialbu: Referencing this here: https://github.com/bane-labs/bridge-evm-contracts/pull/242#discussion_r2182112389

Let's resolve this in the context of the metadata pull request #242

contracts/interfaces/IMessageBridge.sol
```diff
@@ -0,0 +1,30 @@
...
+    function setMessageBridge(uint256 fee, uint256 maxMessageSize, uint256 maxDeposits) external;
+    function messageBridgeIsSet() external view returns (bool);
+    function pauseMessageBridge() external;
+    function unpauseMessageBridge() external;
+    function storeMessage(
+        bytes32 depositRoot,
+        BridgeLib.Signature[] calldata signatures,
+        StorageTypes.MessageData[] calldata messages
+    )
+        external;
```
mialbu, line 26: issue 1: please add `executeMessage()` function as `external` to the interface.

issue 2: format (align the external keyword here to line 25).

contracts/bridge/BridgeImpl.sol
```diff
@@ -581,4 +587,163 @@ contract BridgeImpl is BridgeStorage, IBridge, INativeBridge, ITokenBridge {
...
+        for (uint256 i = 0; i < messageLength; i++) {
+            StorageTypes.MessageData calldata messageData = _messages[i];
+            if (n3ToEvmMessages[messageData.nonce].target != address(0)) revert MessageAlreadyExists(messageData.nonce);
+            // Decode the message bytes into a Call struct and store it directly
+            n3ToEvmMessages[messageData.nonce] = abi.decode(messageData.message, (StorageTypes.Call));
+            emit MessageDeposit(messageData.nonce, messageData.message);
+        }
+    }
+
+    function executeMessage(uint256 nonce) public payable returns (StorageTypes.Result memory) {
```
mialbu: question: why is this public and not external? Do we need to be able to call this function internally?

mialbu (changes requested):
Just 2 minor last things, then we can move forward.

contracts/interfaces/IMessageBridge.sol
OT-rebel: 2. it's interesting that this is what the `forge fmt` does, so unless I fiddle with the formatter, it always does this to me when I commit :(

OT-rebel: 1. done

contracts/bridge/BridgeImpl.sol

contracts/interfaces/IMessageBridge.sol
mialbu: Seems like it's a convention of the foundry formatter then (I get the same behaviour). Let's leave it like it is then.

mialbu (approved):
Looks good to me now to use as a base for further arb-message development on top. 👍

Issue #247, opened 2025-07-10
[AMB] Return Execution Result to N3


mialbu:
Besides returning, we also should consider whether we should allow a `Call` to fail or just return the error message. Context (function `executeMessage`):

```solidity
if (!call.allowFailure && !result.success) revert CallFailed(result.returnData);
```

PR #252, opened 2025-07-18
Send Message from EVM to N3

OT-rebel:


test/MessageBridge.t.sol
```diff
@@ -1088,4 +1088,235 @@ contract MessageBridgeTest is Test, SigUtils {
...
+    }
+
+    function test_SendMessage_MessageTooLarge() public {
+        // Get the message bridge config to know the max size
+        (,,, StorageTypes.MessageConfig memory config) = bridgeProxy.messageBridge();
+
+        // Create a message that is larger than the max allowed size
+        bytes memory largeMessage = new bytes(config.maxMessageSize + 1);
+        for (uint256 i = 0; i < largeMessage.length; i++) {
+            largeMessage[i] = 0xFF; // Fill with non-zero bytes
```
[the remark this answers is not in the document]
OT-rebel: Prefer readability in tests

contracts/messageBridge/MessageBridge.sol
```diff
@@ -112,6 +123,41 @@ contract MessageBridge is IMessageBridge, ReentrancyGuardUpgradeable, UUPSUpgrad
         emit MessageBridgeUnpause();
     }
 
+    /**
+     * @notice Sends a message to the Neo N3 blockchain.
+     * @param _message The message to be sent.
+     */
+    function sendMessage(bytes calldata _message) external payable nonReentrant whenMessageBridgeNotPaused {
```
mialbu, line 130: chore: additional information is needed here.

- should the message be executable or store-only?
- if it is executable, should the result be stored?

We could provide two functions, one for storeOnly, and another for executable messages with an additional boolean parameter for what to do with the result (store or not).

OT-rebel: I would like to address this part in #247

mialbu:
As discussed internally, I'll take over this branch since I've just worked on this code's requirements on the N3 side in https://github.com/bane-labs/bridge-neo-contracts/pull/198.

I'll close this pull request and open a new one, so that Otniel can review later.

PR #260, opened 2025-07-24
Add a sync test for comparison to N3

OT-rebel:


OT-rebel:
@mialbu changes approved!

PR #261, opened 2025-07-24
Send message to N3

mialbu:


contracts/library/MessageBridgeLib.sol
```diff
@@ -41,7 +50,39 @@ library MessageBridgeLib {
...
+        pure
+        returns (bytes32)
+    {
+        return keccak256(abi.encodePacked(_nonce, _version, _sender, _timestamp, _message));
     }
+
+    function _hashMessageSendOp(uint256 _nonce, bytes memory _msgBytes, bytes memory _encodedMetadata) internal pure returns (bytes32) {
+        // Get the top uint32 value to determine the message type
+        // This is used to differentiate between executable, store-only, and result messages
+        uint32 msgType;
```
OT-rebel: the `enum` is actually  encoded as a `uint8` since it can only have up to 256 values.

contracts/library/MessageBridgeLib.sol
```diff
@@ -41,7 +50,39 @@ library MessageBridgeLib {
...
+        bytes memory _message
+    )
+        internal
+        pure
+        returns (bytes32)
+    {
+        return keccak256(abi.encodePacked(_nonce, _version, _sender, _timestamp, _message));
     }
+
+    function _hashMessageSendOp(uint256 _nonce, bytes memory _msgBytes, bytes memory _encodedMetadata) internal pure returns (bytes32) {
```
OT-rebel: I think `nonce`, `_encodedMetadata`, `_msgBytes` is consistent with the order in the `SendMessageData` struct

contracts/library/MessageBridgeLib.sol
mialbu: I'm aware. That's why I convert it to a uint32 when encoding it. But I guess 256 values are plenty even for the future.

contracts/library/MessageBridgeLib.sol
mialbu: I used this order on N3. I was planning to change the order in the struct later and didn't do so yet to not break anything else.

contracts/library/MessageBridgeLib.sol
OT-rebel: lol yeah. if we ever get to 256 types we will probably already be retired

contracts/library/MessageBridgeLib.sol
mialbu: Actually, I guess the order nonce, metadata, then message bytes would make sense though. I'll replicate it in the hash function as well.

contracts/library/MessageBridgeLib.sol
mialbu: Changed in [53a97e2](https://github.com/bane-labs/bridge-evm-contracts/pull/261/commits/53a97e2d4ad6278daf1573e6ced83e83ebb6fd5f)

contracts/library/MessageBridgeLib.sol
mialbu: Changed the order in [53a97e2](https://github.com/bane-labs/bridge-evm-contracts/pull/261/commits/53a97e2d4ad6278daf1573e6ced83e83ebb6fd5f)

OT-rebel (approved):

PR #264, opened 2025-07-28
Add different types of incoming messages

OT-rebel:


contracts/library/MessageBridgeLib.sol
```diff
@@ -47,42 +41,56 @@ library MessageBridgeLib {
     /**
      * @dev Computes the hash of a single message operation.
      * @param _nonce The nonce of the message
-     * @param _message The message content
-     * @return The hash of the message operation
+        * @param _encodedMetadata The encoded metadata of the message
```
mialbu: nitpick: format

contracts/messageBridge/MessageBridge.sol
```diff
@@ -256,9 +259,16 @@ contract MessageBridge is IMessageBridge, ReentrancyGuardUpgradeable, UUPSUpgrad
 
         if (address(messageExecutionManager) == address(0)) revert ExecutionManagerNotSet();
 
+        bytes memory encodedMetadata = storedMessage.encodedMetadata;
+        AMBTypes.MessageType msgType = MessageBridgeLib._readMessageType(encodedMetadata); // Ensure the message type is valid
+
+        AMBTypes.MetadataExecutable memory metadata;
+        if (msgType != AMBTypes.MessageType.EXECUTABLE) revert MessageBridgeLib.UnsupportedMessageType(msgType);
+        else metadata = abi.decode(encodedMetadata, (AMBTypes.MetadataExecutable));
```
mialbu, line 267: nitpick: For the sake of consistency, can you use brackets for if-else statements. I don't mind one-liners for single if statements, but if we add else statements, it can get harder to read with one-liners.

mialbu (approved):
Great! I just have 2 nitpicks. Please consider them but I'll approve already.

contracts/messageBridge/MessageBridge.sol
OT-rebel: unfortunately `forge fmt` does this. I actually wrote it with brackets initially

PR #269, opened 2025-07-28
Send result back

OT-rebel:


test/MessageBridge.t.sol
```diff
@@ -1643,4 +1551,343 @@ contract MessageBridgeTest is Test, SigUtils {
...
+        // Verify second result message was sent
+        StorageTypes.State memory stateAfterSecond = messageBridgeProxy.getMessageBridgeState().evmToN3State;
+        assertEq(stateAfterSecond.nonce, initialState.nonce + 2, "Nonce should be incremented after second result");
+
+        // Verify roots are different
+        assertTrue(initialState.root != stateAfterFirst.root, "Root should change after first result");
+        assertTrue(stateAfterFirst.root != stateAfterSecond.root, "Root should change after second result");
+    }
+
+    function test_SendResultMessage_CanSendSameResultMultipleTimes() public {
```
mialbu, line 1837: I question if we should allow this. It seems like a misuse to me. It's not critical and should only overwrite a result on N3 again with the same data.

mialbu (approved):
Looks good - I added a single comment about a design choice that we still should talk about. It should not stop this pull request from being merged though.

PR #270, opened 2025-07-28
Collect duplicate Message Test Code in Helper

mialbu:


contracts/tests/TestMessageBridge.sol
```diff
@@ -0,0 +1,42 @@
...
+        external
+        override
+        initializer
+    {
+        __ReentrancyGuard_init();
+        _getAMBStorage().management = IBridgeManagement(_management);
+        _setMessageBridge(_fee, _maxMessageSize, _maxNrMessages, _executionWindowSeconds);
+    }
+
+    function hashSendMessage(
```
OT-rebel: I would remove this and just use the MessageBridgeLib directly. 
I think it can get confusing when seeing  something like this in the tests: `messageBridgeProxy.hashSendMessage` because:
a. one could assume that the bridge has a `hashSendMessage` functionality
b. `hashSendMessage` is not really a thing since we hash everything the same way
c. in terms of calls it's 1:1 so there is no reusability benefit

test/MessageBridge.t.sol
```diff
@@ -148,35 +69,6 @@ contract MessageBridgeTest is Test, SigUtils {
...
-        // Extract the boolean (byte at offset 21)
-        // Shift right by 21 bytes (168 bits) to get the boolean at the start
-        // Then mask with 0xff to isolate just that byte
-        // Then check if it's non-zero (Solidity booleans are 1 for true, 0 for false)
-        uint8 executingPausedBoolByte = uint8(uint256(slotValue) >> 168) & 0xFF;
-        return executingPausedBoolByte > 0;
-    }
-
     function test_pauseSending() public {
         assertEq(isSendingPaused(), false, "Sending should not be paused initially");
```
OT-rebel: I'm also wondering if we should just expose the `is*Paused` functionality altogether instead of doing this whole storage reading in the helper?

mialbu: Yes, I was thinking that we should do this.

contracts/tests/TestMessageBridge.sol
mialbu: Done, remove the function and now the test uses the library function directly. Done in [a16144d](https://github.com/bane-labs/bridge-evm-contracts/pull/270/commits/a16144d83df9ebe4c4e4804464dd2086c3661076)

test/MessageBridge.t.sol
mialbu: Done, exposed pause read functions in [3eeee23](https://github.com/bane-labs/bridge-evm-contracts/pull/270/commits/3eeee235c01db45592e22714b33d140e33666fc4).

OT-rebel (approved):

PR #271, opened 2025-07-28
Add expiration time when storing

OT-rebel:


contracts/messageBridge/MessageBridge.sol
```diff
@@ -323,89 +302,115 @@ contract MessageBridge is IMessageBridge, ReentrancyGuardUpgradeable, UUPSUpgrad
...
         emit MessageDeposit(messageData.nonce, messageData.message);
     }
 
+    function _saveAdditionalState(uint256 nonce, bytes memory encodedMetadata) private {
+        AMBStorage.AMB storage ambStorage = AMBStorage.get();
+        AMBTypes.MessageType msgType = MessageBridgeLib._readMessageType(encodedMetadata);
+        if (msgType == AMBTypes.MessageType.EXECUTABLE) {
+            AMBTypes.MetadataExecutable memory metadata = abi.decode(encodedMetadata, (AMBTypes.MetadataExecutable));
+            // Since this is an N3 timestamp, we normalize it to seconds
+            uint256 normalizedTimestamp = metadata.timestamp / 1000;
```
mialbu, line 320: I'm questioning if the expiration should even rely on the initial timestamp. What if the relayer is down for a while due to whatever reasons? Imo, the default expiration should depend on when the message arrives, i.e., the block.time here. It's another case where users want a message to expire sooner to stop it from being executed after a certain time. That would be an additional feature though.

mialbu (approved):

contracts/messageBridge/MessageBridge.sol
mialbu: Let's move with the changes in this pull request and sync about whether to apply my comment above separately.

PR #283, opened 2025-09-02
fix: Change AMBStorage to abstract contract

### LINKED ISSUE #280: [AMB] Hardhat fails to compile MessageBridge because AMBStorage is a library
**Problem**
`AMBStorage.sol` is declared as a library. However, it includes storage-related logic. While it doesn't directly touch storage slots (hence it can be compiled), it defines storage slot usage with the `AMBStorageLocation`. This should not happen in a library. Compiling with hardhat is even failing for this reason, because the NatSpec `/// @custom:storage-location erc7201:AMB.storage` is used (appropriately) on the `AMB` struct.

**Solution**
Change `AMBStorage.sol` to an abstract contract that `MessageBridge.sol` should extend. This would follow the common pattern that is used by OZ's upgradeable contracts. Consider renaming accordingly (e.g., `StorableMessageBridgeUpgradeable`) and as `IMessageBridge` (to force all extending contracts to implement that interface).

OT-rebel:
Fixes #280

### DIFF
--- a/contracts/library/AMBStorage.sol
+++ b/contracts/library/AMBStorage.sol
@@ -5,7 +5,7 @@ import {IBridgeManagement} from "../interfaces/IBridgeManagement.sol";
 import {IExecutionManager} from "../messageBridge/interfaces/IExecutionManager.sol";
 import {StorageTypes} from "./StorageTypes.sol";
 
-library AMBStorage {
+abstract contract AMBStorage {
     //keccak256(abi.encode(uint256(keccak256("AMB.storage")) - 1)) & ~bytes32(uint256(0xff))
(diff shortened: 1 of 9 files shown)


contracts/library/AMBStorage.sol
mialbu, line 1: Please move this file to the `contracts/messageBridge/` directory since it's no longer a library.

mialbu (changes requested):
Just one change requested. Otherwise, looks good!

contracts/library/AMBStorage.sol
OT-rebel: fixed in fbc754a

mialbu (approved):
Great!

PR #296, opened 2025-09-08
feat: Include encoded metadata in MessageSent event (#282)

### LINKED ISSUE #282: [AMB] Include Full Metadata in `MessageSend` Event
**Problem**
In the event emitted when sending messages, only the timestamp and the sender parts of its metadata are included in the event. However, the type and potential other data should also be part of the event to have all data in the logs to reconstruct the hash-chain appropriately.

**Solution**
Replace the `timestamp` in the message sending event with the encoded metadata bytes. The `sender` can be useful to filter for, so it may remain in the event as long as it is indexed.

OT-rebel:
Fixes #282


mialbu (approved):
Looks good! 💯

PR #298, opened 2025-09-10
fix: Address AMB-01 by refunding the transaction value if the execution fails

### LINKED ISSUE #285: AMB-01: Risk of Assets Becoming Stuck `[High]`
A risk of asset lock-up has been identified in the `ExecutionManager` contract when silent failure of external executions is allowed. The `executeMessage` function, marked as `payable`, enables the contract to receive native tokens during cross-chain message execution. However, if the external execution fails and the `allowFailure` flag is set to true, the function does not revert, resulting in the permanent retention of the sent funds within the contract without any recovery mechanism.

### Code refereces
https://github.com/bane-labs/bridge-evm-contracts/blob/f87e5ca5cb1212e085f3d06860002e9f5c87c6ea/contracts/messageBridge/ExecutionManager.sol#L21-L39

https://github.com/bane-labs/bridge-evm-contracts/blob/f87e5ca5cb1212e085f3d06860002e9f5c87c6ea/contracts/messageBridge/MessageBridge.sol#L373

OT-rebel:
Fixes #285

### DIFF
--- a/contracts/messageBridge/ExecutionManager.sol
+++ b/contracts/messageBridge/ExecutionManager.sol
@@ -12,7 +12,7 @@ contract ExecutionManager is IExecutionManager, AccessControl {
     error ExecutionFailed(bytes returnData);
     //0x626ade30
     error ValueMismatch(uint256 providedValue, uint256 expectedValue);
-    //0x0a5b63b0
+    //0xf0c49d44
     error RefundFailed();
 
     constructor(address bridge) {
@@ -31,6 +31,7 @@ contract ExecutionManager is IExecutionManager, AccessControl {
         onlyRole(BRIDGE_ROLE)
         returns (AMBTypes.Result memory result)
     {
+
         // If rawMessage contains data that doesn't match the structure of the Call struct
         // the transaction will fail with a decoding error
         AMBTypes.Call memory call = abi.decode(rawMessage, (AMBTypes.Call));
@@ -41,7 +42,7 @@ contract ExecutionManager is IExecutionManager, AccessControl {
             if (!call.allowFailure) revert ExecutionFailed(returnData);
             // If the call is allowed to fail, refund the value sent for the call (if any)
             if (call.value > 0) {
-                (bool refundSuccess,) = payable(refundAddress).call{value: call.value}("");
+                (bool refundSuccess,) = refundAddress.call{value: call.value}("");
                 // If the refund fails, revert the entire transaction to avoid losing funds
                 if (!refundSuccess) revert RefundFailed();
             }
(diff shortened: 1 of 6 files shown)


contracts/messageBridge/ExecutionManager.sol
```diff
@@ -33,7 +36,16 @@ contract ExecutionManager is IExecutionManager, AccessControl {
         AMBTypes.Call memory call = abi.decode(rawMessage, (AMBTypes.Call));
 
         (bool success, bytes memory returnData) = _executeCall(call.target, call.value, call.callData);
-        if (!success && !call.allowFailure) revert ExecutionFailed(returnData);
+        if (!success) {
+            // If the call failed and it's not allowed to fail, revert the entire transaction
+            if (!call.allowFailure) revert ExecutionFailed(returnData);
+            // If the call is allowed to fail, refund the value sent for the call (if any)
+            if (call.value > 0) {
+                (bool refundSuccess,) = payable(refundAddress).call{value: call.value}("");
```
mialbu: If `refundAddress` is used, it will only be used as payable address. Marking it in the signature is fine and doesn't require the repetitive casting here.

```suggestion
                (bool refundSuccess,) = refundAddress.call{value: call.value}("");
```

test/MessageBridge.t.sol
```diff
@@ -1881,4 +1877,256 @@ contract MessageBridgeTest is MessageBridgeTestHelper {
...
+            })
+        );
+
+        storeMessage(1, messageData, "");
+
+        // Send wrong value - should revert with ValueMismatch
+        vm.prank(eoa);
+        vm.expectRevert(abi.encodeWithSelector(ExecutionManager.ValueMismatch.selector, testValue / 2, testValue));
+        messageBridgeProxy.executeMessage{value: testValue / 2}(1);
+    }
```
mialbu: Please add another message execution transaction using a value greater than the one defined in the messageData. You can add it in the same test function here as the name suggests a value mismatch (cover < and >).

contracts/messageBridge/ExecutionManager.sol
```diff
@@ -12,6 +12,8 @@ contract ExecutionManager is IExecutionManager, AccessControl {
     error ExecutionFailed(bytes returnData);
     //0x626ade30
     error ValueMismatch(uint256 providedValue, uint256 expectedValue);
+    //0x0a5b63b0
```
mialbu:
```suggestion
    //0xf0c49d44
```

mialbu (changes requested):
Well implemented and a great set of tests! 👏🏼 

There's only a few changes I'd like to have. Take a look.

OT-rebel:
@mialbu addressed all in 8f0fc40

contracts/messageBridge/ExecutionManager.sol
mialbu: Fixed in 8f0fc40

test/MessageBridge.t.sol
mialbu: Done in 8f0fc40

contracts/messageBridge/ExecutionManager.sol
mialbu: Fixed in 8f0fc40

mialbu (approved):
Great, LGTM! 👍

PR #300, opened 2025-09-11
feat: Add nonce as return value to send*Message (#287)

### LINKED ISSUE #287: AMB-05: Missing Nonce Return on New Messages `[Info]`
The public methods responsible for sending messages in the EVM bridge implementation (sendMessage and sendExecutableMessage) do not return the nonce associated with the newly generated message. As the nonce serves as a unique and incremental identifier for each message, its immediate availability is critical for tracking message status on the destination network or during processing.

In the current design, the nonce is computed within the internal _sendMessageWithMetadata function but is not returned by the public methods that invoke it. This differs from the NeoVM implementation, where the private updateEvmMessageState method returns the nonce, and it is subsequently exposed by the public methods, allowing users and external systems to immediately access the identifier of the sent message. 

The lack of nonce return in EVM forces integrators to rely on emitted events (MessageSent) or perform additional state queries to retrieve the message identifier, introducing unnecessary complexity and increasing the risk of synchronization issues or traceability loss.

### Code references
https://github.com/bane-labs/bridge-evm-contracts/blob/f87e5ca5cb1212e085f3d06860002e9f5c87c6ea/contracts/messageBridge/MessageBridge.sol#L238-L263

https://github.com/bane-labs/bridge-evm-contracts/blob/f87e5ca5cb1212e085f3d06860002e9f5c87c6ea/contracts/messageBridge/MessageBridge.sol#L157

https://github.com/bane-labs/bridge-evm-contracts/blob/f87e5ca5cb1212e085f3d06860002e9f5c87c6ea/contracts/messageBridge/MessageBridge.sol#L180

https://github.com/bane-labs/bridge-neo-contracts/blob/07f7d0662e34c9ff1f08ad3c84add14a81b739fb/src/main/java/network/bane/message/MessageBridgeImpl.java#L325

OT-rebel:


mialbu (approved):
Looks good 👍🏼

PR #302, opened 2025-09-11
feat: Add getters for AMB storage elements

### LINKED ISSUE #279: [AMB] Add Getters for all entries of the AMB struct
**Problem**
The `AMB` struct is stored using the namespaced storage location standard ERC7201. Its value is accessed using the internal `get` function. It's not accessible externally. There's getters for most of its variables, but not for all of them, i.e., `management`, `executionManager`, and `unclaimedFees`.

**Solution**
There should either be a getter that can return the whole struct or at least all variables of AMB should have a getter.

OT-rebel:


mialbu (approved):

OT-rebel (commented):
LGTM

Issue #277, opened 2025-07-29
[AMB] Add direct lookup for returned N3 method call results


OT-rebel:
See description here: https://github.com/bane-labs/bridge-neo-contracts/issues/207.

mialbu:
This already exists. See here: https://github.com/bane-labs/bridge-evm-contracts/blob/f87e5ca5cb1212e085f3d06860002e9f5c87c6ea/contracts/messageBridge/MessageBridge.sol#L222-L236

mialbu:
I closed this issue too soon. The result I've referred to in the previous comment are results from EVM calls. However, this issue should be about results that were returned from method calls that were executed on Neo N3.

I've updated the title accordingly as the previous title was not clear and concise enough.

mialbu:
This has been implemented in #304

PR #304, opened 2025-09-15
feat: Add direct lookup for returned result messages (#277)

OT-rebel:


contracts/messageBridge/AMBStorage.sol
```diff
@@ -18,6 +18,7 @@ abstract contract AMBStorage {
         mapping(uint256 => StoredMessage) evmMessages;
         mapping(uint256 => bytes) evmExecutionResults;
         mapping(uint256 => ExecutableState) evmExecutableStates;
+        mapping(uint256 => uint256) resultN3NonceToExecutableNonce;
```
mialbu: chore: Please add a public getter in `AMBStorage.sol` as we've done with all other entries in the `AMB` struct.

contracts/messageBridge/MessageBridge.sol
```diff
@@ -225,6 +225,26 @@ contract MessageBridge is IMessageBridge, ReentrancyGuardUpgradeable, UUPSUpgrad
         return abi.decode(message, (AMBTypes.Result));
     }
 
+    /**
+     * @notice Gets the nonce of the result message corresponding to a related message nonce.
+     * @param relatedMessageNonce The nonce of the related message.
+     * @return nonce The nonce of the result message.
+     */
+    function getN3ResultNonce(uint256 relatedMessageNonce) external view returns (uint256) {
+        return getStorage().resultN3NonceToExecutableNonce[relatedMessageNonce];
```
mialbu: chore: In #303 I've started using `getStorage()` outside of `AMBStorage` only when writing to storage is required. Please use the public getter of the `resultN3NonceToExecutableNonce` field I've requested in the previous comment above.

contracts/messageBridge/MessageBridge.sol
```diff
@@ -225,6 +225,26 @@ contract MessageBridge is IMessageBridge, ReentrancyGuardUpgradeable, UUPSUpgrad
...
+    }
+
+    /**
+     * @notice Gets the raw result message for a previously executed message.
+     * @param relatedMessageNonce The nonce of the related message that was executed.
+     * @return result The raw result message bytes.
+     */
+    function getN3Result(uint256 relatedMessageNonce) external view returns (bytes memory) {
+        uint256 resultNonce = getStorage().resultN3NonceToExecutableNonce[relatedMessageNonce];
```
mialbu: chore: same here, please use a public getter instead of accessing the `AMB` storage slot with `getStorage()`.

contracts/messageBridge/MessageBridge.sol
```diff
@@ -329,6 +349,9 @@ contract MessageBridge is IMessageBridge, ReentrancyGuardUpgradeable, UUPSUpgrad
             // in case the relayer is down and does not relay the message in time.
             getStorage().evmExecutableStates[nonce] =
                 ExecutableState({executed: false, expirationTimestamp: block.timestamp + window});
+        } else if (msgType == AMBTypes.MessageType.RESULT) {
+            AMBTypes.MetadataResult memory metadata = abi.decode(encodedMetadata, (AMBTypes.MetadataResult));
+            getStorage().resultN3NonceToExecutableNonce[metadata.relatedMessageNonce] = nonce;
```
mialbu: chore: same here, use public getter instead of `getStorage()`.

test/MessageBridge.t.sol
```diff
@@ -814,7 +814,61 @@ contract MessageBridgeTest is MessageBridgeTestHelper {
...
 
-    // Test depositMessage function with random messages
+    // Test storeMessages function with random messages
+
+    function test_N3ResultFunctions() public {
+        // Step 1: Send an executable message from EVM to N3
+        bytes memory executableMessage = abi.encodePacked("Test executable message to N3");
+
+        // Send the executable message and get its nonce
+        uint256 executableNonce = messageBridgeProxy.sendExecutableMessage{value: messageFee}(executableMessage, true);
```
mialbu, line 831: test: add 2 assertions after this here before storing the result message and linking the result with the related message nonce:

1. `getN3ResultNonce(executableNonce)` should return 0
2. `getN3Result(executableNonce)` should return empty bytes, i.e., bytes with length 0 (same assertion as the one in the last line of this test case).

mialbu (changes requested):
Good work! I have some small change requests before we merge. Please have a look.

contracts/messageBridge/MessageBridge.sol
OT-rebel: this writes to storage

OT-rebel:
@mialbu - addressed in 5cf92d3

contracts/messageBridge/MessageBridge.sol
mialbu: Oh, yes, of course. 😅

PR #312, opened 2025-10-15
feat: Add script to send messages and unpause

OT-rebel:


scripts/messages/sendMessage.ts
```diff
@@ -0,0 +1,129 @@
...
+3. Executable message with custom data:
+   MESSAGE_BRIDGE_ADDRESS=0x... MESSAGE_TYPE=executable MESSAGE_DATA=0x... STORE_RESULT=true npx hardhat run scripts/messages/sendMessage.ts
+
+Environment Variables:
+- MESSAGE_BRIDGE_ADDRESS: (Required) Address of the MessageBridge contract
+- MESSAGE_TYPE: (Required) "executable" or "store-only"
+- MESSAGE_DATA: (Required) Raw hex message data or string
+- STORE_RESULT: "true" or "false" (default: true, for executable only)
+`);
+}
```
mialbu, line 72: That's great to have here! 👏

scripts/messages/sendMessage.ts
```diff
@@ -0,0 +1,129 @@
...
+  try {
+    const config = parseConfig();
+    validateConfig(config);
+
+    console.log("MessageBridge Configuration:");
+    console.log(`- Bridge Address: ${config.messageBridgeAddress}`);
+    console.log(`- Message Type: ${config.messageType}`);
+    console.log(`- Store Result: ${config.storeResult}`);
+
+    const signer = getOwner(ethers.provider);
```
mialbu: chore: please use a different account here. You can also add one, e.g., `getSigner1`, `getDefaultSigner` or something like that from the environment. The owner here is mainly used as "owner" of the management in test deployments. Even though they're test deployments and never actually affect mainnet as the management is and will remain a native contract. I'd prefer to not mix it.

mialbu (approved):
Just nitpicks from my side

PR #317, opened 2025-10-23
feat: Add initialize() to MessageBridge (#315)

### LINKED ISSUE #315: Add initializer to MessageBridge
The MessageBridge is intended to be a native contract on Neo X. Since the genesis is immutable, the MessageBridge will use one of the remaining Stub Proxies. For this - different than with the BridgeContract and the ManagementContract - the initial slot allocation will not be done by setting the slot values in the genesis file but rather through UUPS initialization.

OT-rebel:
Closes #315


contracts/messageBridge/MessageBridge.sol
```diff
@@ -12,16 +12,47 @@ import {ReentrancyGuardUpgradeable} from "@openzeppelin/contracts-upgradeable/ut
...
 
+    function initialize(
+        address _management,
+        uint256 _fee,
+        uint256 _maxMessageSize,
+        uint256 _maxNrMessages,
+        uint256 _executionWindowSeconds
+    )
+        public
+        initializer
```
mialbu: critical: This MUST have the `onlyAdmin` modifier. That's also the reason why we still need the `TestMessageBridge` contract. Please:
- Add the `onlyAdmin` modifier to this initialize function
- Make it virtual (for enabling implementing the next point)
- Add the `TestMessageBridge` again - overriding this `initialize` function without using the onlyAdmin modifier - in order for us to be able to actually initializing it when testing.

mialbu (changes requested):

contracts/messageBridge/MessageBridge.sol
OT-rebel: yeah, good point. I totally missed that. 

however, we only really need the TestMessageBridge for the devnet since in tests we can vm.prank the `GOV_ADMIN`

contracts/messageBridge/MessageBridge.sol
```diff
@@ -12,16 +12,61 @@ import {ReentrancyGuardUpgradeable} from "@openzeppelin/contracts-upgradeable/ut
...
+    function initialize(
+        address _management,
+        uint256 _fee,
+        uint256 _maxMessageSize,
+        uint256 _maxNrMessages,
+        uint256 _executionWindowSeconds
+    )
+        public
+        virtual
+        onlyAdmin
```
mialbu, line 34: chore: Add the `initializer` modifier here (and on the TestMessageBridge's initialize function) and use `onlyInitializing` in the internal `__initialize` function.

contracts/messageBridge/MessageBridge.sol
```diff
@@ -12,16 +12,61 @@ import {ReentrancyGuardUpgradeable} from "@openzeppelin/contracts-upgradeable/ut
...
     }
 
+    function initialize(
+        address _management,
+        uint256 _fee,
+        uint256 _maxMessageSize,
+        uint256 _maxNrMessages,
+        uint256 _executionWindowSeconds
+    )
+        public
```
mialbu: chore: afaik, external should be sufficient. Is there a specific reason you marked it public? If not, please use external.

mialbu (changes requested):

mialbu (approved):
👍

PR #323, opened 2025-11-05
feat: Add script to fund native and token bridges

OT-rebel:
To be used in DevNet

### DIFF
--- a/scripts/fundAllBridges.ts
+++ b/scripts/fundAllBridges.ts
@@ -4,7 +4,8 @@ import {getDeployer, getOwner} from './utils/wallet';
 import { fundAddress } from "./utils/funding";
 
 async function sendETHToBridge() {
-    // Only the owner wallet can send ETH to the bridge because it holds the FUNDER role
+    // Only the account that has the FUNDER role in the bridge can send ETH to it
+    // This function expects the owner to have the FUNDER role (this script is only intended to be used in test envs).
     const owner = getOwner(ethers.provider);
     console.log(`Using owner wallet: ${owner.address}`);
 
--- a/scripts/fundAllBridges.ts
+++ b/scripts/fundAllBridges.ts
@@ -4,7 +4,7 @@ import {getDeployer, getOwner} from './utils/wallet';
 import { fundAddress } from "./utils/funding";
 
 async function sendETHToBridge() {
-    // Get the owner wallet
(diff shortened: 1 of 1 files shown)


scripts/fundAllBridges.ts
```diff
@@ -0,0 +1,72 @@
+import { ethers } from "hardhat";
+import { getBridgeFromEnv } from "./utils/addresses";
+import {getDeployer, getOwner} from './utils/wallet';
+import { fundAddress } from "./utils/funding";
+
+async function sendETHToBridge() {
+    // Get the owner wallet
```
mialbu: chore: please add a comment that only the `funder` role assignee can send ETH to the bridge and that for this script here the owner is assumed to have that role.

scripts/fundAllBridges.ts
```diff
@@ -0,0 +1,72 @@
...
+    const owner = getOwner(ethers.provider);
+    console.log(`Using owner wallet: ${owner.address}`);
+
+    // Get the bridge contract from environment variable
+    const bridge = await getBridgeFromEnv(ethers.provider);
+    const bridgeAddress = await bridge.getAddress();
+    console.log(`Bridge contract address: ${bridgeAddress}`);
+
+    // Amount to send (can be customized via env variable or hardcoded)
+    const amount = process.env.ETH_AMOUNT ? ethers.parseEther(process.env.ETH_AMOUNT) : ethers.parseEther("0.01");
```
mialbu: suggestion: use a number > 1 ETH, e.g., 10 ETH as default as with 0.01 we can't do much anyway.

scripts/fundAllBridges.ts
```diff
@@ -0,0 +1,72 @@
...
+
+    // Get the amount from environment variable
+    const tokenAmount = process.env.TOKEN_AMOUNT;
+    if (!tokenAmount) {
+        throw new Error("TOKEN_AMOUNT environment variable is required for ERC20 transfer");
+    }
+
+    // Get the deployer wallet
+    const deployer = getDeployer(ethers.provider);
+    console.log(`Using owner wallet: ${deployer.address}`);
```
mialbu: fix: there's a mismatch between the log and what wallet is actually read. Stick to either one, deployer or owner.

mialbu (changes requested):

mialbu (approved):

mialbu (approved):

PR #325, opened 2025-11-11
Separate AMB Unpause Scripts and Add Pause Scripts

mialbu:


scripts/utils/pausingHelper.ts
```diff
@@ -0,0 +1,143 @@
...
+      console.log(`No action needed; ${operationName} is already unpaused.`);
+      return {
+        operation: `${operationName} Unpause`,
+        success: true,
+        actionRedundant: true,
+      };
+    } else {
+      console.log(`${operationName} is not paused, proceeding to pause.`);
+    }
+  }
```
OT-rebel: Can be reduced to something like:

```suggestion
  const shouldPause = pausingAction === "pause"
  const isCurrentlyPaused = await isPaused();
  
  const pauseName = shouldPause ? "Pause" : "Unpause";
  const isInDesiredState = shouldPause ? isCurrentlyPaused : !isCurrentlyPaused
  
  if (isInDesiredState) {
    console.log(`${operationName} is already ${pausedAction}d.`);
    return {
      operation: `${operationName} ${pauseName}`,
      success: true,
      actionRedundant: true,
    };
  }
```

mialbu: done in https://github.com/bane-labs/bridge-evm-contracts/pull/325/commits/1ba03529400cc8d83da8b2927ce308333e16e98c

OT-rebel (approved):

PR #336, opened 2026-03-19
[TokenBridge] Return nonce in withdrawal functions

### LINKED ISSUE #335: [TokenBridge] Withdraw functions should return nonce
### Problem

When interacting with the token bridge via other contracts, the nonce from a withdrawal can be important.

Currently, in order to get the nonce of a withdrawal, the bridge contract's withdraw* function needs to be called, and then the contract needs to fetch the current nonce in order to work with the nonce of that withdrawal.

### Solution

Make the withdraw* functions return the nonce similar to how the MessageBridge handles this. This applies to both native and token bridges.

### Reference

https://github.com/bane-labs/bridge-evm-contracts/blob/a70b698be02bda605b2ba1df02248b99574f85b8/contracts/interfaces/INativeBridge.sol#L51

https://github.com/bane-labs/bridge-evm-contracts/blob/a70b698be02bda605b2ba1df02248b99574f85b8/contracts/interfaces/ITokenBridge.sol#L50

https://github.com/bane-labs/bridge-evm-contracts/blob/a70b698be02bda605b2ba1df02248b99574f85b8/contracts/messageBridge/interfaces/IMessageBridge.sol#L42-L50

mialbu:
Closes #335

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -26,5 +26,6 @@ jobs:
       run: npm ci
 
     - name: Run tests
+      env:
+        FOUNDRY_OFFLINE: "false"
       run: npm run test
-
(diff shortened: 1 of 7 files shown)


merl111 (approved):
Straight forward addition.

